PR #10, opened 2025-04-09
Add cooperative matrix 1 (SPV_KHR_cooperative_matrix) proposal

fairywreath:
Adds Cooperative matrix 1 proposal.
* [SPIRV spec](https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_cooperative_matrix.html)
* [GLSL spec](https://github.com/KhronosGroup/GLSL/blob/main/extensions/khr/GLSL_KHR_cooperative_matrix.txt) (for reference). This proposal exactly matches the intrinsics defined here.
* [HLSL Vulkan spec](https://github.com/microsoft/hlsl-specs/blob/dd3931a15c18843313ac5a349735755a6abc6bd0/proposals/0021-vk-coop-matrix.md) (for reference). These are not yet implemented by DXC.

### DIFF
--- a/proposals/023-cooperative-matrix.md
+++ b/proposals/023-cooperative-matrix.md
@@ -1,0 +1,178 @@
+# SP #023: Cooperative Matrix
+
+This proposal introduces support for cooperative matrices in Slang.
+
+## Status
+
+Status: Design Review
+
+Implementation: [PR 6565](https://github.com/shader-slang/slang/pull/6565)
+
+Author: Darren Wihandi
+
+Reviewer: TBD
+
+## Background
+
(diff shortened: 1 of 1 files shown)


csyonghe (commented):
Looks good to me!

csyonghe:
It is probably beyond the scope of this proposal, but the Slang language really need a concept of "local abstract type" that only allows to be used as the type of a local variable, and can't be used as struct fields or declared as members of a storage/constant buffer. Examples are CoopVec, CoopMatrix, RayQuery, HitObject etc.

Right now we are not stating what happens when a user puts `CoopMat` inside a struct `X`, and start to use `X*` or `RWStructuredBuffer<X>`. Such code is not meaningful and should result in an error.

@tangent-vector for viz.

csyonghe (approved):

PR #12, opened 2025-04-25
Cooperative matrix 2 proposal

jkwak-work:


proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,524 @@
...
+  - Runtime states:
+    - permutation[Dim]: Permutation of dimensions (compile-time constant).
+    - viewDimension[Dim]: Logical size for each dimension in the view when HasDimensions is true.
+    - viewStride[Dim]: Stride (in elements) for each dimension in the view.
+    - clipRowOffset: Offset for the row clipping region.
+    - clipRowSpan: Span for the row clipping region.
+    - clipColOffset: Offset for the column clipping region.
+    - clipColSpan: Span for the column clipping region.
+
+CoopMatEnsorView takes (N + 2) number of generic arguments where N is a value given as a first generic argument. As an example, when the first value is 2, CoopMatTensorView must use 4 generic arguments. More examples will be `CoopMatTensorView<1, false, 0>`, `CoopMatTensorView<2, true, 0, 1>` and `CoopMatTensorView<3, false, 0, 2, 1>`.
```
fairywreath, line 88: > 
```suggestion
CoopMatTensorView takes (N + 2) number of generic arguments where N is a value given as a first generic argument. As an example, when the first value is 2, CoopMatTensorView must use 4 generic arguments. More examples will be `CoopMatTensorView<1, false, 0>`, `CoopMatTensorView<2, true, 0, 1>` and `CoopMatTensorView<3, false, 0, 2, 1>`.
```

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,524 @@
...
+
+The first generic argument to `CoopMatTensorView` indicates the dimensionality, and the tailing values indicates the re-ordering of the dimension. A common use cases are `CoopMatTensorView<2, false, 0, 1>` for row-major-matrix load and store, and `CoopMatTensorView<2, false, 1, 0>` for column-major-matrix load and store.
```
fairywreath, line 90:
```suggestion
The first generic argument to `CoopMatTensorView` indicates the dimensionality, and the trailing values indicates the re-ordering of the dimension. A common use cases are `CoopMatTensorView<2, false, 0, 1>` for row-major-matrix load and store, and `CoopMatTensorView<2, false, 1, 0>` for column-major-matrix load and store.
```
Run an LM to check for these kinds of issues, I don't know how to trigger the copilot bot myself.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,524 @@
...
+  - R: The matrix use (see CoopMatMatrixUse), indicating its role in matrix operations
+  - implements IArray<T> and IArithmetic.
+
+Note that a cooperative matrix must be declared with one of three "uses": MatrixA, MatrixB, or MatrixAccumulator. Each operation specialized for cooperative matrices expects specific uses for its operands. For example, a matrix declared as MatrixA cannot be used as MatrixB.
+
+#### CoopMatTensorLayout
+
+While not directly analogous, CoopMatTensorLayout can be loosely compared to the concept of a "subregion" in texture handling. It defines a smaller section of the structure and organization of data within a cooperative matrix tensor.
+
+- CoopMatTensorLayout<Dim, ClampMode>: Describes the layout of a tensor for addressing.
```
fairywreath, line 61: I am OK with this name but I am curious if it is better to drop the `CoopMat` prefix for these tensor types so they can be used for future non-coopmat features(?). This is a snippet from `SPV_NV_tensor_addressing` spec:
```
This extension adds tensor layout and view types which initially can be be used with SPV_NV_cooperative_matrix2. It is written as a separate extension to allow it to potentially be used with other extensions in the future.
```
I am not sure whether there are already other non-coopmat extensions that being planned that use these tensor types.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,524 @@
...
+    let M : int,
+    let N : int,
+    let RM : int,
+    let RN : int,
+    FuncType : IFunc<T, T, T>>
+void coopMatReduce(
+    out CoopMat<T, S, RM, RN, CoopMatMatrixUse::MatrixAccumulator> result,
+    CoopMat<T, S, M, N, CoopMatMatrixUse::MatrixAccumulator> m,
+    CoopMatReduceMask reduceMask,
+    DecodeFuncType combineOp);
```
fairywreath, line 353: I think you mean `FuncType` here.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,524 @@
...
+
+Converts a cooperative matrix to from CoopMatMatrixUse::MatrixAccumulator to CoopMatMatrixUse::MatrixB and transposes the matrix.
+
+```
+__generic<
+    T : __BuiltinArithmeticType,
+    let S : CoopMatScope,
+    let M : int,
+    let N : int>
+void coopMatTransposeNV(
```
fairywreath, line 390:
```suggestion
void coopMatTranspose(
```

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,524 @@
...
+
+__generic<let ClampMode : CoopMatClampMode>
+extension CoopMatTensorLayout<1, ClampMode>
+{
+    void setBlockSize(uint32_t blockSize0);
+    void setStride(uint32_t stride0);
+    void setDimension(uint32_t dim0);
+    void slice(uint32_t offset0, uint32_t span0);
+    void setClampValue(CoopMatClampMode clampMode);
+}
```
fairywreath, line 414: The GLSL and SPIRV intrinsics for these have a builder pattern(function returns a copy with modified internal value) while the functions here are regular setters. I do not really have a strong opinion on which one we should use and I think these are OK.

I looked into examples of existing cm2 shaders such as [this from llama.cpp](https://github.com/ggml-org/llama.cpp/blob/13be08daf992c89d5169518229b3740041c0f419/ggml/src/ggml-vulkan/vulkan-shaders/mul_mm_cm2.comp#L278), and it seems like the builder pattern is  _mostly_ not utilized and there are multiple awkard parts in the code where the builder pattern functions have the same input/output variable. A common usecase for this builder pattern however is for slicing(see occurrences of `sliceTensorLayoutNV` in that file); it seems very convenient to have a slice function that returns a copy directly. Maybe it's worth having a builder pattern function for slicing(or for completion for others as well) alongside these setter extensions.

jkwak-work:
I think Darren brought up good points.
I am open to a suggestion for the name.

And the syntax about the setter is also a good thing to discuss.
Currently the SPIRV functions treat the instances as immutable. The setter functions return new instances with a different value.

The problem with the immutable setters is that it is errorprone.
When you are supposed to do like
```
let newInstance = oldInstance.setStride(32);
```
People often just to like,
```
oldInstance.setStride(32);
```
And there is no warnings or errors about it.
Well... I guess C++ has a way to print a warning for it. But Slang currently doesn't.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,524 @@
...
+```
+__generic<
+    T : __BuiltinArithmeticType,
+    let S : CoopMatScope,
+    let M : int,
+    let N : int,
+    let R : CoopMatMatrixUse,
+    each UserDataTypes,
+    FuncType : IFunc<T, uint32_t, uint32_t, T, expand each UserDataTypes>>
+void coopMatPerElement(
```
csyonghe, line 371: Let's not copy the GLSL design here. Instead of an `out` parameter, let's do
```
CoopMat<T,M,N> coopMatMapElement<T, int M, int N, each U>(
    expand CoopMat<each U, M, N> ms, IFunc<T, expand each U> f);
```

proposals/000-cooperative-matrix-2.md
csyonghe: let's just expose builder pattern here, I think that is easier to use.

jkwak-work:
I am surprised by how quickly I get reviews for this doc considering how long it is.

I appreciate quick reviews!

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+
+
+Related Work
+------------
+
+- GLSL: The GLSL_NV_cooperative_matrix2 extension introduces types like cooperative_matrixNV and a suite of functions for matrix/tensor operations, including variadic template-based APIs for elementwise and reduction operations. The proposed approach in this document is very similar to the GLSL syntax.
+
+- SPIR-V: The extensions SPV_KHR_cooperative_matrix, SPV_NV_cooperative_matrix2, and SPV_NV_tensor_addressing define new opcodes and capabilities for these types and operations.
+
+- Other shading languages: HLSL support is currently limited to cooperative vector features, and cooperative matrix functionality is not supported yet.
```
jeffbolznv, line 32: Note that DXC's SPIR-V backend does have some support for coopmat1 wrapping spir-v intrinsics. See https://github.com/microsoft/DirectXShaderCompiler/pull/6720. It's in a namespace so there shouldn't be any conflicts, but maybe it would make sense to keep syntax aligned where possible?

proposals/000-cooperative-matrix-2.md
jeffbolznv: I agree, the layout is the layout of data in memory and it isn't specifically tied to the coopmat types. So having coopmat in the name is kind of misleading.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+
+CoopMatTensorView takes (N + 2) number of generic arguments where N is a value given as a first generic argument. As an example, when the first value is 2, CoopMatTensorView must use 4 generic arguments. More examples will be `CoopMatTensorView<1, false, 0>`, `CoopMatTensorView<2, true, 0, 1>` and `CoopMatTensorView<3, false, 0, 2, 1>`.
+
+The first generic argument to `CoopMatTensorView` indicates the dimensionality, and the trailing values indicates the re-ordering of the dimension. A common use cases are `CoopMatTensorView<2, false, 0, 1>` for row-major-matrix load and store, and `CoopMatTensorView<2, false, 1, 0>` for column-major-matrix load and store.
+
+Note that CoopMatTensorLayout and CoopMatTensorView are only for loading and storing of CoopMat instances.
+
+### New Enums
+
+- CoopMatScope: Specifies the hardware scope over which the cooperative matrix is distributed.
```
jeffbolznv, line 96: Does slang already have scope enums for synchronization (similar to https://github.com/KhronosGroup/GLSL/blob/main/extensions/khr/GL_KHR_memory_scope_semantics.txt)? If so, it would be good to reuse those. Otherwise, maybe these should also lose the CoopMat prefix.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+
+CoopMat represents a GPU-optimized matrix type enabling distributed, high-performance matrix operations across hardware scopes.
+
+- CoopMat<T, S, M, N, R>: Cooperative matrix type parameterized by element type, scope, dimensions, and matrix use.
+  - T: The element type (e.g., float, int, half)
+  - S: The hardware scope (see CoopMatScope) over which the matrix is distributed
+  - M: The number of rows in the matrix
+  - N: The number of columns in the matrix
+  - R: The matrix use (see CoopMatMatrixUse), indicating its role in matrix operations
+  - implements IArray<T> and IArithmetic.
```
jeffbolznv, line 53: Would IArray include scalar*coopmat multiply?

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+
+- CoopMatTensorLayout<Dim, ClampMode>: Describes the layout of a tensor for addressing.
+  - Dim: The number of dimensions of the tensor layout (e.g., 2 for a 2D tensor).
+  - ClampMode: The clamping or boundary mode for tensor addressing (see CoopMatClampMode).
+  - Runtime states:
+    - blockSize[Dim]: Block size for each dimension.
+    - layoutDimension[Dim]: Logical size for each dimension.
+    - stride[Dim]: Stride (in elements) for each dimension.
+    - offset[Dim]: Offset for each dimension.
+    - span[Dim]: Span for each dimension.
```
jeffbolznv, line 69: You should probably link to the SPIR-V spec for a normative description of what all these values do.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+
+The first generic argument to `CoopMatTensorView` indicates the dimensionality, and the trailing values indicates the re-ordering of the dimension. A common use cases are `CoopMatTensorView<2, false, 0, 1>` for row-major-matrix load and store, and `CoopMatTensorView<2, false, 1, 0>` for column-major-matrix load and store.
+
+Note that CoopMatTensorLayout and CoopMatTensorView are only for loading and storing of CoopMat instances.
+
+### New Enums
+
+- CoopMatScope: Specifies the hardware scope over which the cooperative matrix is distributed.
+  - Device: The entire device (all invocations on the device participate).
+  - Workgroup: The current workgroup (all invocations in the workgroup participate).
```
jeffbolznv, line 98: You should say which scopes are supported, and for Workgroup how does a developer know whether Workgroup is supported (requires a Vulkan coopmat2 feature to be enabled)?

Similarly, how does a developer know whether all the other coopmat2 features are supported? Will slang emulate any of them on coopmat1 implementations?

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+- CoopMatMatrixUse: Indicates the intended use of the matrix in matrix operations.
+  - MatrixA: The matrix is used as the left-hand (A) operand in a matrix multiplication.
+  - MatrixB: The matrix is used as the right-hand (B) operand in a matrix multiplication.
+  - MatrixAccumulator: The matrix is used as the accumulator (result) in a matrix multiplication.
+
+- CoopMatMatrixLayout: Describes the memory layout of the matrix.
+  - RowMajor: Matrix elements are stored in row-major order.
+  - ColumnMajor: Matrix elements are stored in column-major order.
+
+- CoopMatMatrixOperands: Flags that control special behaviors or properties for matrix operations.
```
jeffbolznv, line 111: This shouldn't appear in the high level language. This is because SPIR-V (like LLVM) avoids using a signedness of integer types, and the signedness has to be passed as part of the instruction. But slang does have signed/unsigned integer types. So this is just information that slangc should figure out which generating SPIR-V.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+
+- This setBlockSize(uint32_t blockSize0, ...);
+- This setStride(uint32_t stride0, ...);
+- This setDimension(uint32_t dim0, ...);
+- This slice(uint32_t offset0, uint32_t span0, ...);
+- This setClampValue(CoopMatClampMode clampMode, ...);
+
+The `setBlockSize` function specifies the size of a block, where a block is conceptually similar to a 4x4 block used in texture compression (e.g., DXT).
+
+The `setStride` function defines the number of blocks to the next block in each dimension. For the innermost dimension, the stride will be always 1, and the next outer dimension will have a stride equal to the number of column, and so on.
```
jeffbolznv, line 461: I don't think innermost stride needs to be 1, though that is the common case.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+
+typealias CoopMatType = CoopMat<float, CoopMatScope::Subgroup, 32, 32, CoopMatMatrixUse::MatrixAccumulator>;
+
+[numthreads(32, 1, 1)]
+void computeMain()
+{
+    // Create a 2D tensor layout with clamp-to-edge mode
+    CoopMatTensorLayout<2, CoopMatClampMode::ClampToEdge> tensorLayout;
+    tensorLayout = tensorLayout.setDimension(16, 16);
+    tensorLayout = tensorLayout.setStride(1, 32);
```
jeffbolznv, line 175: The last dimension is the innermost dimension, so this should be (32, 1).

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+```
+
+Detailed Explanation
+--------------------
+
+**Note**: "AnyBuffer" means some or all variants of "RWByteAddressBuffer", "ByteAddressBuffer", "RWStructuredBuffer", "StructuredBuffer", "groupshared array" "array" and/or "pointer".
+
+### Interface implementations of CoopMat
+
+CoopMat implements IArray<T> and IArithmetic. For the required implementation, CoopMat acts as a one-dimensional array. IArithmetic operations will be performed element-wise. Comparing two cooperative matrices will be done lexicographically.
```
jeffbolznv, line 197: Are comparisons necessary? Does IArray or IArithmetic require IComparable?

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+
+### Member functions of CoopMat
+
+Cooperative matrices are accessed as one-dimensional arrays in terms of language syntax. You can retrieve the length of the matrix using the `getLength()` method and access individual elements using the bracket accessor (`[]`) for both reading and writing operations.
+
+```
+void fill(T t);
+void copyFrom<U>(CoopMat<U, S, M, N, R> other)
```
jeffbolznv, line 205: Is there an `operator=`? This would be nicer syntax than copyFrom/fill

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+static int getRowCount();
+static int getColumnCount()
+static uint getLength();
+```
+
+- Load
+```
+static This load(AnyBuffer buffer, uint element, uint stride, CoopMatMatrixLayout matrixLayout);
+static This loadAny<U, let V : int>(__constref groupshared U[V] data, uint element, uint stride,  CoopMatMatrixLayout matrixLayout);
```
jeffbolznv, line 214: Can this also be named `load`?

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+static This load(
+    AnyBuffer buf,
+    uint elementOffset,
+    CoopMatTensorLayout<Dim, ClampMode> tensorLayout);
+
+__generic<
+    let Dim : uint32_t,
+    let ClampMode : CoopMatClampMode,
+    let DimView : uint32_t,
+    let HasDimensions : bool>
```
jeffbolznv, line 240: Do you need the permutation values in here?

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+An example of `decodeFunc` will be,
+```
+T decodeFunc(
+    const T* ptr,  // pointer to the raw data for the element
+    const uint32_t[Dim] blockCoord,   // block coordinates
+    const uint32_t[Dim] coordInBlock  // coordinates within the block
+);
+```
+
+**TODO**: It is unclear if we can use a raw memory pointer for the first parameter.
```
jeffbolznv, line 288: You should also forbid using a decodeFunc with shared memory.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+CoopMat<T, S, M, N, CoopMatMatrixUse::MatrixAccumulator> coopMatMulAdd(
+    CoopMat<U, S, M, K, CoopMatMatrixUse::MatrixA> matA,
+    CoopMat<V, S, K, N, CoopMatMatrixUse::MatrixB> matB,
+    CoopMat<W, S, M, N, CoopMatMatrixUse::MatrixAccumulator> matC,
+    constexpr CoopMatMatrixOperands operands = CoopMatMatrixOperands::None);
+```
+
+### Reduce operation
+
+`coopMatReduce()` computes a matrix where each element of the result matrix is computed from a row, column, or neighborhood of the source matrix depending on `reduceMake`.
```
jeffbolznv, line 338:
```suggestion
`coopMatReduce()` computes a matrix where each element of the result matrix is computed from a row, column, or neighborhood of the source matrix depending on `reduceMask`.
```

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+    FuncType : IFunc<T, uint32_t, uint32_t, expand each U>>
+CoopMat<T, S, M, N, R> coopMatMapElement(
+    expand CoopMat<each U, M, N, R> ms,
+    FuncType mapOp);
+```
+
+`mapOp` is a user-defined function that implements `IFunc<T, uint32_t, uint32_t, U0, U1, U2, ... >`.
+ - The first parameter is the row index (uint32_t row) of the current matrix element.
+ - The second parameter is the column index (uint32_t col) of the current matrix element.
+ - The next parameters (U0, U1, U2, ...) are the corresponding elements from each input cooperative matrix at position (row, col).
```
jeffbolznv, line 379: These can be either cooperative matrices with the same type, or scalars that match the matrix component type.

proposals/000-cooperative-matrix-2.md
jkwak-work: I don't think so.
I think I should change it to implement,
```
IArray<T:IArithmetic> and IArithmetic
```
@fairywreath  for visibility

proposals/000-cooperative-matrix-2.md
jkwak-work: Slang doesn't support `operator=`.
I am not sure about the history but I think this is an explicit decision not to support.

Maybe we should mention about this on our document if it hasn't been, because other Slang users will wonder about it all the time too.

proposals/000-cooperative-matrix-2.md
jkwak-work: @fairywreath named it and I think it was inherited name from how CoopVec does, which @expipiplus1 implemented.
To @expipiplus1 , can you tell us why it got its own name with the postfix "Any" for groupshared?

proposals/000-cooperative-matrix-2.md
jkwak-work: Oh!!! Now that makes more sense.
I will update the document accordingly.

proposals/000-cooperative-matrix-2.md
jkwak-work: Yes, IArithmetic requires IComparable.
```
interface IArithmetic : IComparable
```
I will update the document to explicitly mention IComparable and explain that it is because IArithmetic inherits it.

proposals/000-cooperative-matrix-2.md
jkwak-work: Oh!! I wasn't sure about it.
I will update the document.
Thanks

proposals/000-cooperative-matrix-2.md
jkwak-work: I see. I will change the description so that the innermost dimension can also be other values as well.

proposals/000-cooperative-matrix-2.md
jkwak-work: Thanks for the explanation.
I thought that it must be to workaround a problem something like that but I wasn't sure exactly what.

It sounds like Slang doesn't need this enum at all.
I will remove it.

proposals/000-cooperative-matrix-2.md
jkwak-work: I will check to see if I can reuse any existing enums.
Thanks for pointing this out.

proposals/000-cooperative-matrix-2.md
csyonghe: I see no reason to give it a different name, let's just use `load`.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+ - The first parameter is a pointer to the raw buffer data for the element.
+ - The second parameter is a uint32_t[Dim] array for block coordinates.
+ - The third parameter is a uint32_t[Dim] array for coordinates within the block.
+ - All parameters are read-only.
+ - The function is called for each matrix element, and its return value is stored in the result.
+
+An example of `decodeFunc` will be,
+```
+T decodeFunc(
+    const T* ptr,  // pointer to the raw data for the element
```
csyonghe, line 282: should this be T* or void*?
If it is already T* there is nothing to be decoded.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+__generic<
+    T : __BuiltinArithmeticType,
+    let S : CoopMatScope,
+    let M : int,
+    let N : int,
+    let R : CoopMatMatrixUse,
+    each U,
+    FuncType : IFunc<T, uint32_t, uint32_t, expand each U>>
+CoopMat<T, S, M, N, R> coopMatMapElement(
+    expand CoopMat<each U, M, N, R> ms,
```
csyonghe, line 372: I don't think this will work on our current type system. Variadic arguments must come at the end of the argument list.

Please use the extension Tuple method, that is the only way I've tested that will work for now.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+Converts a cooperative matrix to from CoopMatMatrixUse::MatrixAccumulator to CoopMatMatrixUse::MatrixB and transposes the matrix.
+
+```
+__generic<
+    T : __BuiltinArithmeticType,
+    let S : CoopMatScope,
+    let M : int,
+    let N : int>
+void coopMatTranspose(
+    out CoopMat<T, S, N, M, CoopMatMatrixUse::MatrixB> result,
```
csyonghe, line 403: These functions should just return a CoopMat, don't return via `out`.

csyonghe: Similar to all other functions.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+__generic<
+    let Dim : uint32_t,
+    let ClampMode : CoopMatClampMode = CoopMatClampMode::Undefined>
+struct CoopMatTensorLayout
+{};
+
+__generic<let ClampMode : CoopMatClampMode>
+extension CoopMatTensorLayout<1, ClampMode>
+{
+    This setBlockSize(uint32_t blockSize0);
```
csyonghe, line 421: remove all the `set` prefix here.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+{
+... implement ...
+};
+
+__generic<
+    let HasDimensions : bool,
+    let p0 : uint32_t,
+    let p1 : uint32_t,
+    ...>
+extension CoopMatTensorView<..., HasDimensions, p0, p1, ...>
```
csyonghe, line 514: We can choose to support maximum to 6 dimensions for now.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+__generic<
+    T : __BuiltinFloatingPointType,
+    let S : CoopMatScope,
+    let M : int,
+    let N : int,
+    let RM : int,
+    let RN : int,
+    FuncType : IFunc<T, T, T>>
+void coopMatReduce(
+    out CoopMat<T, S, RM, RN, CoopMatMatrixUse::MatrixAccumulator> result,
```
csyonghe, line 350: prefer return value over `out` parameter.

jeffbolznv: The main reason to use an out parameter is to implicitly determine some template parameters. Up to you whether you think that's worth it.

proposals/000-cooperative-matrix-2.md
jeffbolznv: The SPIR-V extension only supports 5 dimensions.

proposals/000-cooperative-matrix-2.md
csyonghe: In this case though, the result type can be deduced from the FuncType, so we should also be fine here.

jeffbolznv: The number of rows/columns in the result matrix is arbitrary (at least for row/col reductions), so it can't be deduced.

csyonghe: I think we should have separate functions to do row/col/2x2 reduction, so the result type can be deduced.

Otherwise, this will open up chances for the result type to be inconsistent with the reduction mode, which will lead to crash/undefined runtime behavior.

proposals/000-cooperative-matrix-2.md
jkwak-work: Good catch. I will change it to `void`.

proposals/000-cooperative-matrix-2.md
jkwak-work: I will change as suggested.

proposals/000-cooperative-matrix-2.md
jkwak-work: I will go with 6 for now.
I am aware the SPIRV spec said 5 is max.

csyonghe: If spirv only supports 5, let's just support up to 5.

proposals/000-cooperative-matrix-2.md
jkwak-work: I will add dedicated functions for each deduction type.
It will allow me to remove the CoopMatReduceMast enum too.

proposals/000-cooperative-matrix-2.md
csyonghe: Also I wonder if this should just be a member on CoopMatrix type?

proposals/000-cooperative-matrix-2.md
csyonghe: same: should this be a member function instead?

proposals/000-cooperative-matrix-2.md
jkwak-work: When I tried, it compiled but I guess it still wouldn't work.

I will change it to what you suggested,
```
CoopMat<T,M,N> coopMatMapElement<T, int M, int N, each U>(
    expand CoopMat<each U, M, N> ms, IFunc<T, int, int, expand each U> f);
```

proposals/000-cooperative-matrix-2.md
jkwak-work: Right... I will make them member functions.

proposals/000-cooperative-matrix-2.md
jkwak-work: Ok. I will rename them.

proposals/000-cooperative-matrix-2.md
jkwak-work: Yes, I will make it a member function.

proposals/000-cooperative-matrix-2.md
jeffbolznv: Even if you have separate functions for row/col/2x2, you still can't deduce the matrix dimensions for row/col reductions because they can be semi-arbitrary. The reduced value is splatted across that dimension. For example, if I have an NxM matrix and do a row reduction, the result can be for example Nx16, and won't be Nx1 because that's not a shape that the implementation supports.

proposals/000-cooperative-matrix-2.md
jeffbolznv: I don't think void is the right thing here, and should maybe even be forbidden. The addressing calculations use the size of the pointed to structure, see: https://github.com/KhronosGroup/GLSL/blob/main/extensions/nv/GLSL_NV_cooperative_matrix2.txt#L519

csyonghe: It should be a different generic parameter then.

```
T decodeFunc<U>(
    const U* ptr, 
```

We currently cannot forbid `void` and other abstract types from the type system. That will need to come with a future type system cleanup around the different kinds of types such as non-copyable, unsized etc.

csyonghe: Also note that we don't support const pointers yet, so const U* is meaningless right now.

proposals/000-cooperative-matrix-2.md
csyonghe: hmm... I see. So `Reduce` really has a fused semantic of splatting.

Let's make it a member function, separate out the functions, remove the ReductionDirection enum, and define only necessary generic params for the reduced dimensions. For example, a row-reduction does not need RN, and col-reduction does not need RM.

Let's use the return value instead of `out` parameter for now. This will mean the user has to explicitly specify the result dimension in the angle bracket today. However with the direction the Slang language is moving, we are going to find a way to add bidirectional type checking so the generic argument can be deduced from the expected return type in the future, and that will allow the user to skip the explicit argument down the road with the return value style intrinsic.

proposals/000-cooperative-matrix-2.md
csyonghe: `IArray` currently only represents a 1-dimensional array. What does it mean to do `coopMat[0]`? Does it return a new coopMat? I am OK if we decide to have CoopMat not conform to `IArray` for now.

jkwak-work: So.. this is related to GLSL syntax for the cooperative matrix 1.
It says,
>     The components of a cooperative matrix are spread across the invocations
>    in its scope, in an implementation-dependent manner. The components owned
>    by a given invocation can be accessed using array subscripting syntax,
>    and the number of components owned by each invocation can be queried
>    using the *length* method. The type returned by *length* is an int, and
>    the value returned is a constant expression. There is no compile-time
>    bounds checking of array indices.
>
>    This can be used, for example, to perform component-wise operations on
>    all components of a cooperative matrix:
>
>      fcoopmatNV<16, gl_ScopeSubgroup, 16, 8> m;
>      ...
>      for (int i = 0; i < m.length(); ++i) {
>          m[i] = f(m[i]);
>      }
>
>    The arithmetic binary operators add (+), subtract (-), and divide (/)
>    operate on cooperative matrix types and perform the operation
>    component-wise.
https://github.com/KhronosGroup/GLSL/blob/main/extensions/nv/GLSL_NV_cooperative_matrix.txt

proposals/000-cooperative-matrix-2.md
jkwak-work: I noticed that HLSL takes the layout value as a template argument and the function names start with a capital letter.

I think that the following functions can be renamed to match to HLSL function names:
- getLength() -> GetLength();
- load(AnyBuffer, int elemOffset, int stride, CoopMatMatrixLayout) -> Load<let layout:CoopMatMatrixLayout>(AnyBuffer, int elemOffset, int stride);
- store(AnyBuffer, int elemOffset, int stride, CoopMatMatrixLayout) -> Store<let layout:CoopMatMatrixLayout>(AnyBuffer, int elemOffset, int stride);

proposals/000-cooperative-matrix-2.md
jkwak-work: I will use `U` as suggested so that it can be any struct or type.

proposals/000-cooperative-matrix-2.md
csyonghe: Ah I see, then we do need IArray here.

proposals/000-cooperative-matrix-2.md
csyonghe: I think you want:
```
struct CoopMat<...> {
    CoopMat<T,M,N> mapElements<T, int M, int N>(
        IFunc<T, int, int, T> f);  // one matrix case.
}

extension<each T, int M, int N> Tuple<expand CoopMat<each T, M, N>> {
     CoopMat<TR, M, N> mapElements<TR>(IFunc<TR,  int, int, expand each T> f);
}
```

This allows users to write:
```
CoopMat<float, ...> m1, m2, m3;
let r1 = m1.mapElements((int x, int y, float e) => ... );
let r2 = (m1, m2, m3).mapElements((int x, int y, float a, float b, float c) => ...);
```

Which is nice and consistent.

proposals/000-cooperative-matrix-2.md
jkwak-work: It appears that Slang doesn't have enum for the memory scope yet.
I will need to add the following,
```
enum MemoryScope
{
    CrossDevice, // Scope crosses multiple devices.
    Device,      // Scope is the current device.
    Workgroup,   // Scope is the current workgroup.
    Subgroup,    // Scope is the current subgroup.
    Invocation,  // Scope is the current Invocation.
    QueueFamily  // Scope is the current queue family.
};
```
https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_scope_id

proposals/000-cooperative-matrix-2.md
jkwak-work: Right. The diagram and explanation on SPIRV doc is very good. I will simplify the description here and add a link to it.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+static This load(
+    AnyBuffer buf,
+    uint elementOffset,
+    CoopMatTensorLayout<Dim, ClampMode> tensorLayout,
+    CoopMatTensorView<DimView, HasDimensions> tensorView);
+
+__generic<
+    let Dim : uint32_t,
+    let ClampMode : CoopMatClampMode,
+    FuncType : IFunc<T, T*, uint32_t[Dim], uint32_t[Dim]>>
```
csyonghe, line 250: I think this may need to be an ordinary `functype` and not an `IFunc`, because `IFunc` can have a captured context, but SPIRV requires this function to be just an ordinary function with no additional `this` parameter.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,536 @@
...
+    let M : int,
+    let N : int,
+    let RM : int,
+    let RN : int,
+    FuncType : IFunc<T, T, T>>
+void coopMatReduce(
+    out CoopMat<T, S, RM, RN, CoopMatMatrixUse::MatrixAccumulator> result,
+    CoopMat<T, S, M, N, CoopMatMatrixUse::MatrixAccumulator> m,
+    CoopMatReduceMask reduceMask,
+    FuncType combineOp);
```
csyonghe, line 353: this should be `functype(T,T)->T combineOp` because SPIRV doesn't allow `this` in a reduce function.

proposals/000-cooperative-matrix-2.md
jkwak-work: We plan to remove the requirement of IComparable from IArithmetic.
So I will keep it as it is for now.

proposals/000-cooperative-matrix-2.md
jkwak-work: Oh! That's what you meant.
I will change to it.

proposals/000-cooperative-matrix-2.md
jkwak-work: I renames as following:
- CoopMatTensorLayout -> TensorLayout
- CoopMatTensorView -> TensorView

proposals/000-cooperative-matrix-2.md
jkwak-work: It seems like you are right. The SPIRV doesn't say we can carry a userData along with the function pointer,
> For OpCooperativeMatrixLoadTensorNV instructions with a DecodeFunc operand, rather than loading a value, the function operand is invoked for each matrix element at least once. The function’s return type must match the component type of the result matrix type. The first parameter must be a pointer type with storage class PhysicalStorageBuffer, and the parameter is filled a pointer computed by multiplying the index returned by matrixCoordToTensorElement(WithView) by the size of the pointee type. The second and third parameters must each be an array of 32-bit integers whose dimension matches the tensor dimension. The second parameter is filled with the blockCoord, and the third parameter with the coordInBlock, for the matrix element being decoded. The return value is stored in the corresponding element of the result matrix.

I am not sure what you meant by "ordinary functype" tho.
The function signature is currently changed to the following,
```
    __generic<
        U,
        let Dim : uint32_t,
        let ClampMode : CoopMatClampMode>
    static This Load(
        AnyBuffer buf,
        uint elementOffset,
        TensorLayout<Dim, ClampMode> tensorLayout,
        IFunc<T, const U*, uint32_t[Dim], uint32_t[Dim]> decodeFunc);
```
Should I change to something like this??
```
    __generic<
        U,
        let Dim : uint32_t,
        let ClampMode : CoopMatClampMode,
        FuncType>
    static This Load(
        AnyBuffer buf,
        uint elementOffset,
        TensorLayout<Dim, ClampMode> tensorLayout,
        FuncType decodeFunc);
```

csyonghe: It shall be:
```
__generic<
        U,
        let Dim : uint32_t,
        let ClampMode : CoopMatClampMode>
    static This Load(
        AnyBuffer buf,
        uint elementOffset,
        TensorLayout<Dim, ClampMode> tensorLayout,
        functype(const U*, uint32_t[Dim], uint32_t[Dim])->T decodeFunc);
```

jkwak-work: What??? I have never seen this syntax. lol...

proposals/000-cooperative-matrix-2.md
jkwak-work: All matrices must have a same element type.
But I think `each T` will allow them to have different types.
```
extension<each T, int M, int N> Tuple<expand CoopMat<each T, M, N>> {
     CoopMat<TR, M, N> mapElements<TR>(IFunc<TR,  int, int, expand each T> f);
}
```
I am not sure how to make the types to be all same.

csyonghe: You can do 
`where U==T`

csyonghe: I don’t think our type system can express this. So unfortunately this will need to be a emit time error.

csyonghe: One thing to try:

extension<T1, each T, int M, int N> Tuple<CoopMat<T1, M, N>, expand CoopMat<each T, M, N>> where T==T1 {
     CoopMat<T1, M, N> mapElements<T1>(IFunc<T1,  int, int, expand each T> f);
}

proposals/000-cooperative-matrix-2.md
jeffbolznv: Right, in SPIR-V cooperative matrices are treated like composites. You can extract/insert elements as if it were a one-dimensional vector/array.

proposals/000-cooperative-matrix-2.md
jkwak-work: I see.. I can try that.
If it doesn't work out, I will force it with `static_assert`.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,552 @@
...
+  - Repeat: Repeat the tensor data (wrap around).
+  - RepeatMirrored: Repeat the tensor data with mirroring at the boundaries.
+
+### Hello World Example 1
+```
+ByteAddressBuffer inputC; // { 1.f, 1.f, 1.f, ... }
+RWStructuredBuffer<float> outputBuffer;
+
+typealias CoopMatAType = CoopMat<float16_t, CoopMatScope::Subgroup, 16, 16, CoopMatMatrixUse::MatrixA>;
+typealias CoopMatBType = CoopMat<float16_t, CoopMatScope::Subgroup, 16, 16, CoopMatMatrixUse::MatrixB>;
```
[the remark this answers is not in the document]
jkwak-work: Oh. I see. I will update the doc with `.`

csyonghe:
The revision looks good. One minor suggestion: can we change to Reduce2x2 for readability?

jkwak-work:
I found that the name `TensorView` is already in use by diff.meta.slang.
https://github.com/shader-slang/slang/blob/dde654c8605ea65f959fcadaa6618337ab1f20c0/source/slang/diff.meta.slang#L206

It seems like the type was introduced to support CUDA side functionality.
I will have to come up a different name for this.

jkwak-work:
> I found that the name `TensorView` is already in use by diff.meta.slang. https://github.com/shader-slang/slang/blob/dde654c8605ea65f959fcadaa6618337ab1f20c0/source/slang/diff.meta.slang#L206
> 
> It seems like the type was introduced to support CUDA side functionality. I will have to come up a different name for this.

I will go with TensorLayoutView.

csyonghe:
TensorLayourView is confusing. Just use TensorView.

We can wrap all of the CoopMat and Tensor types into a namespace linalg, similar to what HLSL is doing.

jkwak-work:
> TensorLayourView is confusing. Just use TensorView.
> 
> We can wrap all of the CoopMat and Tensor types into a namespace linalg, similar to what HLSL is doing.

I see.
I will put them all in a namespace, `linalg`.

jkwak-work:
I pushed a new update just now with the following changes.

The following information get update on the proposal doc.
1. `ReduceTwoByTwo()` is renamed to `Reduce2x2()`
2. Added new capabilities.
3. Added `[require]` lines on all function signatures.
4. Added multiply-with-scalar operator.
5. Mention HLSL side progress.
6. Mention that CoopMat will be placed in a namespace `linalg`
7. `::` is replaced with `.` as many as possible.
8. Fixed out-dated function names in the example codes.
9. `const` keyword is removed from `const U*`, because it causes
   compiler to crash.

proposals/000-cooperative-matrix-2.md
jkwak-work: I think I dismissed this comment prematurely.
The initial concern was about multiply-with-scalar.

The latest doc now includes two operators that works with a scalar value.
```
[require(cooperative_matrix)]
__generic<
    T : __BuiltinArithmeticType,
    let S : MemoryScope,
    let M : int,
    let N : int,
    let R : CoopMatMatrixUse>
CoopMat<T, S, M, N, R> operator *(CoopMat<T, S, M, N, R> lhs, const T rhs);

[require(cooperative_matrix)]
__generic<
    T : __BuiltinArithmeticType,
    let S : MemoryScope,
    let M : int,
    let N : int,
    let R : CoopMatMatrixUse>
CoopMat<T, S, M, N, R> operator *(const T lhs, CoopMat<T, S, M, N, R> rhs);
```

proposals/000-cooperative-matrix-2.md
jkwak-work: It turned out that the current implementation we have with `copyFrom` is mainly to convert the type.
I am not sure what would be a nicer syntax for copying but I think we can add it later when we come up something.

proposals/000-cooperative-matrix-2.md
jkwak-work: It turned out that if I do the following, the comma operator returns just `m3` and it doesn't form a Tuple we expected.
```
let r2 = (m1, m2, m3).mapElements((int x, int y, float a, float b, float c) => ...);
```
The following seems to work,
```
typealias CoopMatType = CoopMat<float, 16, 16, ...>;
let r2 = Tuple<CoopMatType, CoopMatType, CoopMatType>(m1, m2, m3).mapElements((int x, int y, float a, float b, float c) => ...);
```
I am probably gonna need a helper function similar to,
```
let r2 = makeTuple(m1, m2, m3).mapElements((int x, int y, float a, float b, float c) => ...);
```

csyonghe: This parenthesis syntax hasn’t been implemented yet, we can easily add this to the parser.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,671 @@
+SP #021: Cooperative Matrix and Tensor Types
+=======================
+
+Status
+------
+
+Status: Design Review
+
+Implementations:
+ - cooperative_matrix is completed with d0b6a0b1ab49b5958015f31364c5ad73d9cd03eb
```
csyonghe, line 10: link to the PR, if possible.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,671 @@
...
+ - cooperative_matrix_2 is in prototype stage
+
+Author: Jae Hyuk Kwak
+
+Reviewer: TBD
```
csyonghe, line 15: You can update this field with people who reviewed this PR so far.

proposals/000-cooperative-matrix-2.md
```diff
@@ -0,0 +1,671 @@
...
+
+As explained earlier, TensorView takes (N + 2) number of generic arguments where N is a value used as the first generic argument. We could make use of variadic generic value if Slang supports it, but it will be still difficult to define the relationship between a value to the number of generic arguments.
+
+To achieve the same syntax as GLSL does, we will use `extension` feature with a fixed number of maximum dimension value. TensorView will be declared with a fixed number of generic argument. The unwanted arguments will have a default value.
+
+```
+__generic<
+    let Dim : uint32_t,
+    let HasDimensions : bool,
+    let p0 : uint32_t = 0xff,
```
csyonghe, line 602: I feel like 0 is the more appropriate default value.

jkwak-work: Let me change it to zero.

jkwak-work: Hmmm. I think it will be 0xff for the following reason,
When an intended type is, `TensorView<2, false, 0, 1>`, the user may have a mistake like, `TensorView<2, false, 0>`.
It will be a valid value when the default is 0 as `TensorView<2, false, 0, 0>`.
Whe the default value is an invalid value like 0xff or any value bigger than `Dim`, we can catch a mistake like this because it will be `TensorView<2, false, 0, 0xff>`

proposals/023-cooperative-matrix.md
```diff
@@ -12,6 +12,11 @@ Author: Darren Wihandi
 
 Reviewer: TBD
 
+Update:
```
csyonghe, line 15: Can you update the reviewer status here?

Also please review the SP numbers and make sure they don't clash with existing proposals.

PR #13, opened 2025-04-28
Add lambda expression proposal.

csyonghe:


proposals/025-lambda-1.md
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+    let lam = _slang_Lambda_test_0(c);
+    int d = lam.operator()(2);
+}
+```
+
+## Restrictions
+
+Lambda expression in this proposed version can only read captured variables, but not modify them.
```
tangent-vector, line 111: It would be good to also have the restriction that when a lambda expression captures a mutable variable (or more generally an "abstract storage location," once we get around to specifying what that means...), it is an error for other code to mutate that variable during the lifetime of the lambda.

To try to be more precise:

* For each abstract storage location that a lambda closes over, there is an access operation to that storage, starting at the point where the lambda gets created, and ending at whatever point the lifetime of the lambda ends.

* As per our existing hand-waved rules for accesses (that cover things like `inout` parameters, etc.), a program is invalid if there are two overlapping accesses to the same abstract storage location, if at least one of those accesses is a write.

(To be clear: it should be possible for the compiler to statically detect and diagnose many cases where storage accesses would overlap in invalid ways, but for the language we have it would be impossible to provide full static checking, so we are stuck with the hard cases being undefined behavior...)

proposals/025-lambda-1.md
```diff
@@ -0,0 +1,141 @@
...
+    int c = 0;
+    let lam = (int x) {
+        c = c + 1; // Error: c is read-only here.
+        return x + c;
+    };
+    int d = lam(3);
+}
+```
+
+We plan to allow mutating captured variables in a future proposal.
```
tangent-vector, line 126: Avoiding capture of mutable variables also means we are avoiding engaging with the issue of lifetimes/scoping/escaping, but it might be good to set some conservative ground rules for now.

I would like to propose that at least for now:

* It is invalid for a lambda (the result of evaluating a lambda expression) to outlive the scope in which it was created (we'd need to get precise about the details, but informally I think we all understand what this implies).

* For a function that receives any `IFunc` parameters, it is invalid if the function body does anything that would make the argument function value "escape" (outlive the invocation of that function).

We can relax those restrictions as needed over time, but being as conservative as possible for now could save us some semantic headaches that arise from rushing head-first into some thorny semantic challenges.

proposals/025-lambda-1.md
```diff
@@ -0,0 +1,141 @@
...
+
+Lambda expression is not allowed to have mutable parameters, such as `inout` or `out` parameters in this version.
+
+A variable whose type is `[NonCopyable]` cannot be captured in a lambda expression.
```
tangent-vector, line 130: It's good to make a note of this restriction.

Altogether, this proposal makes it clear that, whether we like it or not, Slang has no real option but to start taking ownership, lifetimes, borrows, etc. more seriously.

proposals/025-lambda-1.md
```diff
@@ -0,0 +1,141 @@
...
+
+# Conclusion
+
+This proposal adds limited support for lambda expressions that cannot mutate its captured environment.
+Although being limited in functionality, this kind of lambda expressions will still be very useful
+in many scenarios including the cooperative-matrix operations.
```
tangent-vector, line 136: One thing I might have expected to see (although I haven't looked at the relevant SPIR-V extensions, so maybe they can handle something like a closure...) is a distinction being made in the type system between "stateless" closures (more or less equivalent to just a function pointer, with the given signature) vs. "stateful" ones (more or less some hidden storage of type `T` and a pointer to a function that has the given signature plus an extra `T` parameter).

There's a lot of nuance there around how to handle calling conventions and so forth, and around avoiding generating extra "thunk" values when, e.g., passing a plain old function to an operation that accepts/expects a closure).

tangent-vector (approved):
This is a great proposal to see. It is very encouraging when cutting-edge GPU capabilities are helping to drive the push for Slang to adopt more and more "grown-up" language features, like closures/lambdas.

proposals/025-lambda-1.md
csyonghe: We could potentially express this as:
```
interface IStaticFunc<T, each TArgs> {
    static T operator()(expand each TArgs args);
}
```
But this seems like it is worth having a separate proposal for.

proposals/025-lambda-1.md
csyonghe: Added to restrictions section.

proposals/025-lambda-1.md
csyonghe: Added to restrictions section.

proposals/025-lambda-1.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+The compiler will translate it into:
+
+```slang
+void test()
+{
+    int c = 0;
+    struct _slang_Lambda_test_0 : IFunc<int, int> {
+        int c;
```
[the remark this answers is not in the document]
csyonghe: Discussed in more detail. Clarified that the entire object is captured.

proposals/025-lambda-1.md
```diff
@@ -0,0 +1,141 @@
...
+
+Lambda expressions can be used in positions that accepts an `IFunc`:
+
+```
+void apply(IFunc<int, float> x) {...}
+
+void test()
+{
+    apply((float x)=>(int)x+1); // OK, passing lambda to `IFunc<int, float>`.
+}
```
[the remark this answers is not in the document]
csyonghe: Added to restrictions section.

proposals/025-lambda-1.md
```diff
@@ -0,0 +1,141 @@
...
+        return x;
+    } else {
+        return 0;
+    }
+}
+```
+
+A lambda expression will evaluate to an annoymous struct type that implements the `IFunc` interface
+during type checking. The return type of the lambda function is determined by the body expression (in the case
+of the lambda expression contains a simple expression body), or the value of the first return statement in the
```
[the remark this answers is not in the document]
csyonghe: That makes sense, changed the proposal to require all of them being the same for now.

proposals/025-lambda-1.md
```diff
@@ -0,0 +1,141 @@
...
+(int x) => {
+    if (x > 0) {
+        return x;
+    } else {
+        return 0;
+    }
+}
+```
+
+A lambda expression will evaluate to an annoymous struct type that implements the `IFunc` interface
```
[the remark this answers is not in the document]
csyonghe: added discussion, this will be left for a future extension.

tangent-vector (approved):
Thank you for making those edits.

For anybody following from the sidelines, this proposal is a good example of how to cover the major issues, without going into *too* much detail, for a proposal where the implementation is still ongoing.

proposals/025-lambda-1.md
```diff
@@ -0,0 +1,224 @@
+# SP #025: Lambda Expressions (Immutable Capture)
+
+This proposal adds initial support for lambda expressions in Slang.
+The initial proposal is to support immutable capture of variables in the surrounding scope.
+This means that the lambda can only read the values of the captured variables, not modify them.
```
jeffbolznv, line 5: If there is a functional or language interaction reason for this restriction it would be nice to explain it. Or if it's just to reduce the scope of the initial implementation, that's also good to know.

proposals/025-lambda-1.md
```diff
@@ -0,0 +1,224 @@
...
+## Background
+
+SP009 introduced `IFunc` interface to represent callable objects. This allowed Slang code to
+pass around functions as first-class values by defining types that implement `IFunc`.
+However, this approach is not very convenient for users, as it requires defining a new type for each function
+that needs to be passed around.
+
+This problem can be solved with lambda expressions, which enables the compiler to synthesize such
+boilerplate types automatically. The recent cooperative matrix 2 SPIRV extension introduced several opcodes
+such as Reduce, PerElement, Decode etc. that can expressed naturally with lambda expressions.
```
jeffbolznv, line 26:
```suggestion
such as Reduce, PerElement, Decode etc. that can be expressed naturally with lambda expressions.
```

proposals/025-lambda-1.md
```diff
@@ -0,0 +1,224 @@
...
+
+## Proposal
+
+The proposal is to add the following syntax for lambda expressions:
```
jeffbolznv, line 30: Are all accessible values captured? Is there any way to write a "pure" lambda?

proposals/025-lambda-1.md
```diff
@@ -0,0 +1,224 @@
...
+```slang
+void test()
+{
+    Composite c = {};
+    struct _slang_Lambda_test_0 : IFunc<int, int> {
+        // Lambda captures the entire object instead of just
+        // `member1`.
+        Composite c;
+        __init(Composite in_c) {
+            c = in_c;
```
jeffbolznv, line 142: Are all types copyable/assignable in slang? Or if it's not, would trying to use its value in the lambda lead to a compile error?

proposals/025-lambda-1.md
csyonghe: There is just to keep the initial scope small. There isn't a fundamental reason why we can't allow mutable captures, especially when targeting SPIRV/Metal where pointers are available.

proposals/025-lambda-1.md
csyonghe: Values are captured on-demand. Unreferenced variables are not captured. To write a "pure" lambda, just make sure the lambda doesn't reference any local variables or parameters in the outer function.

proposals/025-lambda-1.md
csyonghe: `RayQuery` and `HitObject` types are currently not copyable, and use of these in the capture will lead to a compile error.

proposals/025-lambda-1.md
```diff
@@ -0,0 +1,224 @@
...
+
+```slang
+struct Composite
+{
+    int member1;
+    float member2;
+    void apply()
+    {
+        let lam = (int x)=>{
+            return x + member1; // captures the entire `this`.
```
Devon7925, line 163: Is there a reason why the whole object has to be captured? Seems like this could lead to users accidentally making expensive copies when working with large structures.

proposals/025-lambda-1.md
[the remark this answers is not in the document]
juliusikkala: I'm interested in interaction with [the error handling proposal.](https://github.com/shader-slang/spec/blob/main/proposals/legacy/003-error-handling.md) Compatibility with `IFunc` already enforces that `throws` cannot be used. Would it make sense to mention in restrictions that lambdas do not support throwing exceptions for now?

proposals/025-lambda-1.md
csyonghe: The decision to capture whole object ensures semantic consistency, so there is never confusion on situations such as:
```
struct Obj { int a; int b; };
int ar[10];
Obj o;
var lam = () =>
{
     f(o);   // captures o, obviously
     g(o.a);  // do we capture another copy of o.a or reuse o?
     ar[0]; // capture 0-th element or entire array?
     ar[i]; // what do we do?
};
```

By saying we always capture the root object at all access chains, none of above ambiguous/difficult cases will ever arise.

The redundant context can be cleaned up very easily in an IR pass post initial IR lowering. Spirv-Opt and most drivers are capable to clean this up today, so it is unlikely going to cause severe performance problems.

If performance is ever a concern on some extreme edge cases, the user can always store the value into a local variable first, and then refer to that local variable within the lambda.

proposals/025-lambda-1.md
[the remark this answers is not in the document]
csyonghe: Yes, this will be an inherited restriction from `IFunc`. We can mention it here for completeness.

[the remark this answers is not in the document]
csyonghe: Added the clarification.

PR #14, opened 2025-05-13
Type System for Compile-Time and Run-Time Existential Values

csyonghe:


tangent-vector (approved):
This proposal looks great. I've walked through the plan as documented here and it seems like all the points are covered and it should be implementable on top of the current compiler codebase (rather than requiring any significant rearchitecture or overhauls).

This would be a really nice feature to get into the language ASAP, so that we can give our users (and ourselves) more confidence in the robustness of code that uses interface-typed parameters/variables/etc.

proposals/024-any-dyn-types.md
```diff
@@ -0,0 +1,250 @@
...
+
+If a variable, a parameter or a struct field is declared to have an interface type or array of interface type, the var decl must be qualified with either the `some` keyword or the `dyn` keyword. If the var decl is not explicitly qualified with `some` or `dyn`, the compiler implicitly qualifieds the type as `some` when language version is 2026 or later, or as `dyn` in language version 2025.
+
+The concrete type of a `some` qualified var decl must be compile-time deducible to a proper type and will never involve in dynamic dispatch. The concrete type of a `dyn` qualified var decl might not be compile-time deducible and might involve in dynamic dispatch. 
+
+The following rules apply to `some` and `dyn` qualifiers on var decls:
+
+1. A `some IFoo` type is valid only if `IFoo` is an interface type.
+1. A `some IFoo` type can only be used on a function parameter, a function return value or a local variable. Struct fields or global variables cannot have a `some` type.
+1. A `some IFoo` type cannot appear in any complex type expression. For example, `some IFoo*`, `some IFoo[2]`, `Tuple<some IFoo, some IFoo>` etc. are not allowed.
```
juliusikkala, line 66: Is there a deeper reason for this limitation, or is it ruled out to keep implementation complexity in check? These seem like they could often be useful.

juliusikkala (commented):
I like this proposal a lot, I've already encountered a few situations where these would've been helpful. I especially appreciate making `some` the default behaviour for 2026.

proposals/024-any-dyn-types.md
```diff
@@ -0,0 +1,250 @@
...
+    float compute(int x);
+}
+```
+
+By default, a `dyn` interface is subject to the following restrictions, unless language version is 2025 or `-enable-experimental-dynamic-dispatch` flag is present:
+
+1. The interface itself cannnot be generic.
+1. It must not define any associated types.
+1. It must not define any generic methods.
+1. It must not define any mutating methods.
```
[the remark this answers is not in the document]
csyonghe: A lot of limitations here are just because our compiler hasn’t implemented such use yet, and can be relaxed later.

proposals/024-any-dyn-types.md
```diff
@@ -0,0 +1,250 @@
...
+The following rules apply to `some` and `dyn` qualifiers on var decls:
+
+1. A `some IFoo` type is valid only if `IFoo` is an interface type.
+1. A `some IFoo` type can only be used on a function parameter, a function return value or a local variable. Struct fields or global variables cannot have a `some` type.
+1. A `some IFoo` type cannot appear in any complex type expression. For example, `some IFoo*`, `some IFoo[2]`, `Tuple<some IFoo, some IFoo>` etc. are not allowed.
+1. A local variable of `some IFoo` type cannot be assigned or initialized more than once throughout its lifetime.
+1. If a function returns `some IFoo`, then all return statements must return values of exactly the same type.
+1. If a function contains a `out some IFoo` parameter, then all values assignmened to the parameter must have exactly the same type.
+1. Two different var decls of `some IFoo` type are considered to have **different** types.
+1. A `dyn IFoo` type is valid only if `IFoo` is an interface type with `dyn` qualifier.
```
ArielG-NV, line 71: This line reads a bit confusingly (give the example provided later in the doc `dyn IFoo d = f;` and `dyn IFoo v1;`). 

For clarification, this means that given the following:
```
dyn interface IDyn{};
some interface ISome{};
struct ISomeImpl : ISome {};
```
1. I cannot do (in a function's scope):
`dyn ISome dynVar;`

2. I can do  (in a function's scope):
`dyn IDyn dynVar;`

3. I can do  (in a function's scope):
```
some ISome someVar = ISomeImpl();
dyn ISome dynVar = someVar;
```
4. I can do (in a function's scope):
```
dyn ISome dynVar = ISomeImpl();
```

proposals/024-any-dyn-types.md
```diff
@@ -0,0 +1,250 @@
...
+1. It must not define any associated types.
+1. It must not define any generic methods.
+1. It must not define any mutating methods.
+1. It cannot contain any methods that has a `some IFoo` return type, or has any `some IFoo` parameters.
+1. A `dyn` interface cannot inherit from any interfaces that are not `dyn`.
+1. A `dyn` interface cannot contain any function requirements that are marked as `[Differentiable]`.
+
+Any type that conforms to one or more `dyn` interface is subject to these restrictions:
+
+1. The type itself must be an ordinary data type, meaning that it cannot contain any fields that are opaque or non-copyable or unsized.
```
ArielG-NV, line 49: Is this section intentionally omitting `unless language version is 2025 or `-enable-experimental-dynamic-dispatch` flag is present:`?
(I ask since this PR is a breaking change without the clause)

kaizhangNV: I don't think we allow opaque type in an interface that can be used in dynamic dispatch for now.
So I guess this is `intentionally omitting`.

proposals/024-any-dyn-types.md
kaizhangNV: Do you have answer for this? Because I have the same question.

proposals/024-any-dyn-types.md
kaizhangNV: Actually, I think my memory is correct. Look at this line.

> Another big down side is that our dynamic dispatch logic can only handle ordinary data types. If a type implementing an interface contains opaque types such as `Texture2D`, the dynamic dispatch pass will fail and give user confusing error messages

proposals/024-any-dyn-types.md
kaizhangNV: I don't think `some` can be used on interface type, from the proposal, `some` is supposed to be used only on var decl.

proposals/024-any-dyn-types.md
csyonghe: No, it is simply because our implementation isn’t handling this case well and we should lift this restriction when we complete our implementation.

proposals/024-any-dyn-types.md
csyonghe: Yes, this applies in all cases regardless of language version or experiemental flag setting. We can never support opaque types in dynamic dispatch.

proposals/024-any-dyn-types.md
csyonghe: The understanding is correct, although I would comment that IInterface() is never a valid expression. You can construct an interface typed value using the interface type.

PR #15, opened 2025-05-19
C++ Modules proposal

expipiplus1:


csyonghe:
The proposals in the spec doc is intended for language design proposals only. Compiler implementation related proposals can just be an issue or discussion on the slang repo.

PR #17, opened 2025-05-21
Update error handling proposal to reflect current state

juliusikkala:
Update error handling proposal to match the current state of [the error handling PR.](https://github.com/shader-slang/slang/pull/6916)

I suggest these changes:
* `{ catch(){} }` -> `do{}catch{}`
* Catch-all when `catch` has no parameter
* Turn `throw` into a statement to avoid needing the `Never` type (should be easy to relax into an expression once that exists)
* Signature change using `Result<T, E>` instead of `Optional<E>` + extra output param
* Remove subtyping considerations for now, as they seem a bit outdated with `IFunc` et al. playing the higher-order function role these days and there doesn't seem to be a clear way to allow throwing from an `IFunc`.

I also updated implementation details and fixed some typos.

I left `IError` as a remaining divergence from the current state of the PR, so either it should be removed from here or the PR should add it back again. I do understand the reasoning here for `IError`, but I wonder if it's actually necessary? You can use the catch-all syntax instead of catching `IError`, and users can create their own interfaces if they are so inclined.

### DIFF
--- a/proposals/026-error-handling.md
+++ b/proposals/026-error-handling.md
@@ -1,4 +1,4 @@
-Error Handling
+SP #026: Error Handling
 ==============
 
 Slang should support a modern system for functions to signal, propagate, and handle errors.
@@ -6,7 +6,13 @@ Slang should support a modern system for functions to signal, propagate, and han
 Status
 ------
 
-In discussion.
+Status: In Experiment.
+
+Implementation: [PR 6619](https://github.com/shader-slang/slang/pull/6619) [PR 6916](https://github.com/shader-slang/slang/pull/6916)
+
+Author: Theresa Foley, Julius Ikkala
+
+Reviewer: TBD
 
 Background
 ----------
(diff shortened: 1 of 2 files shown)


csyonghe:
Let's move this proposal out of `legacy` folder, and give it a proper proposal number just like other proposals. Put your name on the proposal too!

csyonghe:
Also make sure to add links to the PRs implementing `defer` and `catch`, and update the status to `In Experiment`.

csyonghe:
Let's make a copy of this proposal in the `proposals` directory, and remove any inconsistent design/statements (e.g. `IError`) that isn't being implemented.

juliusikkala:
Someone else authored most of this before me, so I guess their name / names should be on it too. Based on previous discussions in the implementing PR, I assume that it's at least @tangent-vector?

csyonghe:
Yes, you can put Theresa Foley there.

PR #20, opened 2025-05-28
Add yet another proposal on initialization

tangent-vector:


proposals/026-initialization-in-slang-2026.md
```diff
@@ -0,0 +1,444 @@
...
+The use of `default` expressions will be the preferred/idiomatic way to declare that a `struct` field should have a default value that matches a default-initialized value of its type.
+
+In order to facilitate use of `default` expressions in generic contexts, the core module will define an interface:
+
+    interface IDefaultable
+    {
+        __init();
+    }
+
+Types can opt-in to support for this interface, but the compiler will *not* automatically treat all types that support a zero-argument constructor call as conforming to `IDefaultable`; it is an explicit `interface`, like any other.
```
tdavidovicNV, line 359: I don't think I understand this. Is this just a syntactic sugar? What is the difference between:
```
struct Foo
{
    var material0 = MyComplicatedMaterialType();
    var material1 : MyComplicatedMaterialType == default;
}

void main()
{
    var f = Foo();
}
```
Will both `material0` and `material1` get their default constructor called?
Can I do:
```
void main()
{
    var material0 = MyComplicatedMaterialType();
    var material1 : MyComplicatedMaterialType == default;
}
```
and get the same result just without having them wrapped in `Foo` struct?

proposals/026-initialization-in-slang-2026.md
```diff
@@ -0,0 +1,444 @@
...
+    Wave w;
+    w.amplitude = 100.0;
+    w.phase = 0.0;
+
+In this example, the attempt to assign to `w.amplitude` will be diagnosed as an error, because `w` is being accessed when it is uninitialized.
+
+Disallowing such a commonly-used idiom may turn out to be something that user's get upset about, or even cannot tolerate.
+However, it is important to note that as soon as one tries to support this kind of idiomatic C-like code, special-case rules and hacks need to creep their way into the language semantics in an attempt to make that code Just Work as users intend.
+
+If we have any aspiration to drop the hacky mess of approaches we have today, we pretty much *have* to abandon support for this idiom.
```
tdavidovicNV, line 379: Can we do:
```
var w = Wave();
w.amplitude = 100.0;
w.phase = 0.0;
```
If so, we will probably mandate that as the semantic for compability with the large C++-adjusted user base.

We will need guidance what to do if the struct contains an array of expensive-to-initialize types. Consider:
```
struct BsdfStack
{
    IBSDF bsdf_stack[8]; // can be about 100B each
    float3 bsdf_weight[8]; // we do not care about the values here, would we do the var bsdf_weight[8] : float3 = default syntax? Definitely do not want Optional here
    int next_bsdf = 0;
}
```
which is then constructed by a large generated code that uses functions like:
```
void add_bsdf(inout BsdfStack stack)
{
    stack.bsdf_stack[stack.next_bsdf++] = ...;
}
```
I assume we will have to do `Optional<IBSDF>` here, even though we know that once the large initialization is done, we can lift the optionality?

Would then have:
```
struct BsdfStackTemp
{
    Optional<IBSDF> bsdf_stack[8];
}
// and
struct BsdfStackFinal
{
    IBSDF bsdf_stack[8];
    __init(BsdfStackTemp temp)
    {
        for (int i = 0; i < 8; ++8)
            bsdf_stack[i] = temp.bsdf_stack[i];
    }
}
```
?

proposals/026-initialization-in-slang-2026.md
```diff
@@ -0,0 +1,444 @@
...
+
+### Constructors Calls and Casts
+
+We propose to keep the current special-case rule where `Thing(a)` is treated as a synonym for `(Thing) a` and first tries built-in type coercion rules before falling back to a constructor call.
+There are plenty of valid reasons why code might end up, e.g., casting a value to its own type (it can easily happen as code evolves), and our past experience has been that *not* treating `Thing(a)` and `(Thing) a` as semantically equivalent ends up confusing users.
```
tdavidovicNV, line 384: Referencing https://github.com/shader-slang/slang/issues/6959 what happens when `Foo` has `__init(Bar b)` and you just assign `var f = Foo(b)` is clear (tries to cast, if it fails, goes looking for the constructor). But what will happen when I do: `Foo f = b;`? Will it fail and require explicit `Foo f = Foo(b)`?

proposals/026-initialization-in-slang-2026.md
```diff
@@ -0,0 +1,444 @@
...
+
+### Initializer Lists
+
+We propose that `{}`-enclosed initializer list syntax should still be supported, at least for the forseeable future, even in Slang 2026.
+However, unlike the situation in current Slang, an initializer-list expression `{a,b,c}` that gets coerced to type `Thing` will be entirely semantically equivalent to `Thing(a,b,c)`.
+We will drop support for the C-style initializer list handling rules, and the special-case rules for forcing zero-initialization even on fields that didn't specify a default value.
+
+The empty initializer list `{}` will be semantically equivalent to a `default` expression.
+
+While initializer lists will be *allowed*, the Slang tools, documentation, etc. should default to showing examples that use only constructor-based initialization or `default` expressions (which should be explained as a shorthand for a zero-argument constructor call).
```
tdavidovicNV, line 394: I think this explains my previous question, but just to get it clear clear:
```
Foo f = {};
Foo f = default;
Foo f = Foo();
var f = Foo();
```
are all equivalent and valid in all the same contexts?

proposals/026-initialization-in-slang-2026.md
```diff
@@ -0,0 +1,444 @@
...
+#### Eventually: Dynamically-Sized Arrays
+
+Another convience feature is probably long-overdue for addition to the core module is some kind of dynamically-sized array type with a fixed upper bound.
+With such a type, a user could write, e.g.:
+
+    BoundedSizeDynamicArray<int, 10> a; // or whatever the type gets called...
+    for(int i = 0; i < 10; i++)
+        a.add( someFunc(i) );
+
+The implementation of such a type in the core module would be largely straightforward, and the core module could take care of whatever delicate interactions it needs to have with the compiler to reassure it that values are consistently initialized before they are accessed.
```
tdavidovicNV, line 437: The question of converting from the BoundedSizeDynamicArray to a static array remains, but I guess you can just assign it.

proposals/026-initialization-in-slang-2026.md
```diff
@@ -0,0 +1,444 @@
...
+
+Alternatives Considered
+-----------------------
+
+The only real alternative we have considered so far is to give up, and resign ourselves from slowly sliding into the realm that C++ occupies where almost all types support being implicitly default-constructed, but often to a state that is literally or figuratively equivalent to a null pointer.
+
+"When every variable is 'intialized', none of them are."
```
tdavidovicNV, line 444: I do not see a problem with variables being initialized to `uninitilized` value. It is pretty much the same as you are proposing after all. The objects must be in some valid state by the time you use them in any way imaginable. And whether that is implicit, or explicit is mostly a matter of taste.

tdavidovicNV (commented):
The reactions are written in the order in which I read the document, so some nitty gritty I thought of before I got to the nitty gritty section.. feel totally free to close things as duplicate or answer multiple in a single place or whatever makes sense.

In general, I like it. I really like the *named arguments*. I worry about the arrays a lot. I did initially miss that this effectively completely bans incremental construction, and I am quite sure that will lead to things just being zero initialized all the time, and then overwritten. 

Which I do not see as a big problem (it would be my take on this, tbh), but it very much leads to the C++ way you don't like too much. But it might be a good compromise, because it mostly won't be too expensive to work The C++ Way, and if you decide to go a more optimal route (less writes after writes), the compiler will check it.

I'd rate this 8/10 on the scale of "if you implement it, I will try it", with couple caveats outlined in the comments, mostly the dynamically sized array of expensive types, and structs containing interfaces.

proposals/026-initialization-in-slang-2026.md
```diff
@@ -0,0 +1,444 @@
...
+
+(Similarly, any attempt to return from a the body of a function-like declaration when one of its `out` parameters is uninitialized is an error)
+
+Notable case of expression that are an error when applied to an uninitialized variable:
+
+* Field/property access, even when on the left-hand side of an assignment. The statement `v.x = 0;` is invalid if `v` is not initialized.
+
+* Method calls on the variable (`v.doThing()`), including to `[mutating]` methods that might otherwise write valid values into the variable.
+
+* Array subscripting (`v[i]`), even when used to initialize the elements of an array.
```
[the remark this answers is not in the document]
tangent-vector: Under the proposal as written, you would *either* need to change your declaration to:

    Thing v[8] = default;

and then you can set the elements one-by-one, in whatever order you like...

... *or* you would need to assign to the entire array in one go, with something like:

    v = Array( (int i) => whatever(i) );

using one of the utility functions being hand-wavily proposed for addition to the `Array` type.

The first of those options only works if `Thing` supports initialization with `default`, and the second requires changing the structure of the code away from what is familiar to most of user current users.

It might be important to expand the related work section of this proposal to include a more complete survey of the idioms that are used to handle arrays in practice in other modern languages... but I fear that the answer in those cases will be that use of dynamically-sized arrays is pervasive, so you'd handle most of these cases by creating `v` as an empty array and then adding elements to it, rather than by creating an array with a non-zero number of elements (but where all the elements are uninitialized), and then overwriting them.

proposals/026-initialization-in-slang-2026.md
```diff
@@ -0,0 +1,444 @@
...
+
+2. A call `__init(a, b, c, ...);` to another constructor of the same type.
+
+3. Zero or mor statements that are allowed to access `this` as a fully-initialized value
+
+The body of a designated constructor must conform to the following sequence of operations:
+
+1. Zero or more statements that may assign to fields of the current type through `this`, but may not otherwise access `this` (notably: cannot access derived fields).
+  At the end of these statements, all fields of the current type that do not have a default-value expression must be fully initialized.
+  (The rest will then be initialized to their default values)
```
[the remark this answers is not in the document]
tangent-vector: I understand you on that, and it initially *seems* like we could obvious stuff the default values in at the start, based on what we do/don't see people set manually:

    struct Foo
    {
        int a = expensiveFunction(0);
        int b = expensiveFunction(1);
        int c = expensiveFunction(2);

        __init(int x, int y)
        {
            this.b = x;
            this.c = y;
        }
    }

In that case a user might feel like it is "obvious" that the compiler should initialize `this.a` to `expensiveFunction(0)` at the start of their constructor, before the rest of the computation runs.

> For a C++ programmer, it might seem like intuitively the compiler would initialize *all three* of `a`, `b`, and `c` at the start of that constructor... which is why I made them be calls to a hypothetical expensive function in my example, to show why that doesn't feel like a good solution.
> 
> C++ works around this problem by inventing the separate list of field/base initializers that run before the constructor body. If you think about it for a bit, it might be clear that C++ has the same basic steps (1), (2), (3), with (3) being the actual body statements of the constructor, and (1) + (2) being handled via the initializers. So what we are proposing to do here is to instead write out parts (1) and (2) as ordinary code, rather than have an entirely separate syntax with all kinds of corner cases and special rules.
> 
> (We also swap around the order of execution of steps (1) and (2). C++ does (2) first, and then (1), which turns out to be kind of obviously wrong, given how it means that your code in step (3) is still very restricted in what it can safely do with `this` (for fear of calling through to a pure virtual function, etc.))
> 
> Anyway...

If you accept that we don't want to initialize all three of `a`, `b`, `c` to their defaults before the user code starts running, you are left with only the idea that the compiler should infer that it must initialize `a`, and do so before the user code runs...

...however, it's entirely possible that the user code might *conditionally* initialize `this.a`:

        __init(int x, int y)
        {
            this.b = f(x);
            this.c = g(y);
            int z = h(x,y);
            if(z > 0) this.a = z;
        }

In this case, we can't determine statically if the user-authored constructor body will initialize `this.a` or not, until after we've actually *run* the user-authored code for step (1). I've written the code to use function calls here to make it clear that we can't just hoist the evaluation of `h(x,y)` to above the rest of the code, to determine whether or not `this.a` gets initialized in the user code or not; doing so could alter the semantics of the code if `f`, `g`, and `h` have observable side effects. (Also imagine that `expensiveFunction()` has side effects too... or at least it *could*)

Thus we have to contend with the fact that either:

* We would have to accept that `this.a` might only be initialized sometimes, and we can only determine at run-time whether the compiler needs to initialize it *after* the user's code has already run. Thus for consistency it only makes sense to evaluate all such field initial-value expressions *after* the user code for part (1) has run.

* Alternatively, we could make it an error for the code in part (1) to conditionally initialize a field of `this`, and thus insist that the code in part (1) must either initialize a given field along all control-flow paths, or none of them. Then we could move the initialization of the fields that statically *don't* get initialized during (1) to the top, and everything maybe works like the user expects. This option sounds reasonable on paper, but it would quickly fall into the trap where users expect the compiler to see/know that "obviously" their code that is under an `if(0)` (or the Slang equivalent of an `if constexpr (0)`...) shouldn't be considered when deciding what the code in part (1) does or does not initialize... and then everything gets muddy again.

* A final alternative is to declare that the order in which default field initial values are computed is unspecified, and it is also unspecified whether they run before or after the user code in part (1). That option is pretty nice, and actually opens up some interesting paths (e.g., you could maybe imagine letting the initial value for one field depend on the value of another, and let the compiler sort out the order of evaluation...), but it doesn't avoid the reality that no matter how "unspecified" the order is, there will be user code that accidentally ends up relying on whatever order happens to fall out of the compiler implementation, and those users will be understandably upset by any/all compiler changes that silently alter the meaning of their code. In practice, if we leave this "unspecified" in theory, we would still want to codify what the compiler *actually* does, and commit to not changing that behavior without careful communication with users.

With all of those options weighed, I think the simplest in terms of complexity for both users and the compiler is the first option: any fields that the user doesn't initialize in part (1) get initialized to their default values by the compiler in between parts (1) and (2).

proposals/026-initialization-in-slang-2026.md
tangent-vector: All of your code above leads to `f.material0`, `f.material1`, `material0`, and `material1` all being initialized with what amounts to a call `MyComplicatedMaterialType()`.

The `default` syntax is, as proposed, just a bit of sugar around calling a zero-argument constructor.
What is most important about it is:

1. It avoids repeating the name of the type being initialized, in contexts where the type should be obvious. There is no need to ever write `ReallyLongTypeName myVar = ReallyLongTypeName();`

2. While a `default` expression means basically the same thing that people use `{}` for today, it is significantly more explicit about the programmer's intention.

The proposed `IDefaultable` interface is a somewhat different matter; the point of the interface is to allow people to write `= default` for types that come from generic parameters. Without it you'd have a problem like:

    void myFunc<T : IWhatever>(...)
    {
        T a[10]; // uninitialized
        a[0].setThing(...); // ERROR!
    }

If you are in a context where you know/expect that the things that will get passed to `myFunc` should always support default initialization, then you can write:

    void myFunc<T : IWhatever & IDefaultable>(...)
    {
        T a[10] = default;
        a[0].setThing(...);
    }

and it will all Just Work. The same is true for generic `struct`s that hold a field of some generic parameter type `T`, but want to have a zero-argument constructor.

Yes, this is kind of just contributing to a slippery slope toward pervasive "initialization" of all variables/fields, even if the values they are initialized to aren't meaningful. I may not personally want to *encourage* programming like that, but I also don't want the language to be (too) actively antagonistic to that style.

If somebody likes doing pervasive default-"initialization" of everything, then they can just remember to make all their types inherit `IDefaultable` and to declare all their variables/fields with `= default`. They may feel that it's pure busy-work to have to do it that way, but it means other programmers aren't forced to adopt the same style, and indeed those very same programmers may find over time that they don't actually need to rely on `= default` as heavily as they thought they did.

(Making all types auto-inherit something like `IDefaultable` when they can, and making all variables/fields do the equiavalent of `= default` when they can, together would make it nearly impossible to ever change away from that behavior, either for individual users or for the language at large.)

proposals/026-initialization-in-slang-2026.md
tangent-vector: I was curious before when I saw that `int next_bsdf` field, and this comment confirmed for me that the `BsdfStack` type ends up functioning as a kind of bounded-but-dynamically-sized array.

Ignoring the SOA-vs-AOS layout issue for just a second, there doesn't seem to me to be any reason this type couldn't be:

```
struct BsdfStack
{
    struct Entry
    {
        IBSDF bsdf;
        float3 weight = default;
    }
    BoundedDynamicArray<8, Entry> entries;

    [mutating] void addBSDF( IBSDF b, float3 w = 0 )
    {
        entries.add( Entry(bsdf: b, weight: w) );
    }
}
```

The data that would be generated for a `BoundedDynamicArray<N, T>` would be the same as for a `T[N]` array plus an `int count` field, and then the code that's generated for the `.add` operation wouldn't be any different in behavior from your `stack.bsdf_stack[stack.next_bsdf++] = ...` logic.

AFAICT the only major reason why you wouldn't be able to adopt that directly is the AOS-vs-SOA thing, and there's a strong argument to be made that a good language in this space should provide some kind of built-in utility type(s) for expressing an SOA-layout array, so ideally you'd just use the composition of those features.

Even without that, I would argue that a `BoundedDynamicArray<N, T>` is both more convenient and more explicit for your use case than an `Array<N, Optional<T>>` would be, because it encodes more of the invariants that actually exist.

It would be simple enough, I expect, to support a built-in method on `BoundedDynamicArray<N,T>` called something like `getValueAssumingAllElementsFilled()`, or something more concise, that returns an `Array<N,T>` and encodes that the user is assuming that the count of the dynamic array is equal to `N` in that case.

proposals/026-initialization-in-slang-2026.md
tangent-vector: Your understanding is correct. To make it clear, because the language documentation might actually not be careful about this (although the spec *will* be...): there are two different scenarios in which the Slang compiler has a value/expression of type `FromType` and is trying to convert it to a value of type `ToType`:

* *Explicit conversion* is triggered specific when the user does `ToType(f)` or `(ToType) f` (both are equivalent) on an expression `f` of type `FromType`.

* *Implicit coercion* is used in contexts where there is no apparent cast or constructor call. That includes cases like `ToType x = f;`, or a function call `foo(f)` to some function `void foo(ToType x)`. This is also what gets used when you do something like `if( f ) ...` ; the expression gets implicitly coerced to `bool` (its like any other implicit coercion where the `ToType` is `bool`).

Both of these operations bottleneck through the same logic in the compiler, and will consider the same *kinds* of operations in the same order. E.g., they will both do the "are `FromType` and `ToType` the same?" check before moving on to other things.

Both cases will eventually bottom out in looking for a constructor on `ToType` that can be invoked with an argument of `FromType`.
At that point the two cases differ in the following very specific ways:

* In the case of implicit coercion, if the search for a usable constructor on `ToType` succeeds, but that constructor is not marked as usable for implicit coercion operations (currently only the core module ever defines such constructors...), then an error will be diagnosed, noting that implicit coercion is now possible between those types, but an *explicit* conversion would be allowed.

* In the case of implicit coercion, the compiler will *not* consider a candidate constructor on `ToType` with a parameter of some type `IntermediateType` such that implicit coercion of `FromType` to `IntermediateType` would require invoking a constructor in `IntermediateType`. This constraint is both to avoid the possibility of an infinite regress in the compiler (trying to convert `A -> C` via intermediate type `B`, and then finding that it can only convert `A -> B` via intermediate type `C`, and then...) and because there is a general sense that users would be reluctant to accept the compiler inserting multiple implicit constructor calls behind their back.

To bring this all back to the concrete example:

    Foo f1 = Foo(b); // allowed: explicit conversion from Bar->Foo
    Foo f2 = b; // not allowed: implicit coercion from Bar->Foo

proposals/026-initialization-in-slang-2026.md
tangent-vector: Those would all be equivalent if we implement this proposal as written.

I believe it is currently *not* the case in Slang that those are all equivalent (ignoring that we don't support `default` at all). The handling of initializer lists and constructors is different code, and while an initializer list might be translated into a constructor call, it is not true that there is an easy and obvious equivalence.

Even if it turns out all of those are equivalent today for an empty argument list, it is definitely not the case that all of those options are equivalent for anything where there are one or more arguments (e.g., `{a, b, c}` vs. `Foo(a,b,c)`).

I would consider one of the main goals of this proposal to be fixing that problem, so that initializer lists are either eliminated (unlikely) or are always equivalent to a constructor call with the same arguments.

tangent-vector: The follow-up here is that I personally think there may be cases where it might make sense to have `Foo f = default;` be allowed even when `Foo f = Foo();` wouldn't be, but I am explicitly *not* including any of that complexity in the proposal, because my goal at this point was to try and write a sketch for The Simplest Thing That Could Possibly Work, rather than prematurely add in all the bells and whistles we might want in practice.

proposals/026-initialization-in-slang-2026.md
tangent-vector: Yes, we could either make a function in the core module that handles the conversion for you, or a user could just write it for themself if we give them some of the higher-order-function ways of creating an ordinary array:

    BoundedSizeDynamicArray<T, N> dynArray = default;
    // ... fill in `dynArray`

    let array = Array<T, N>( (int i) => dynArray[i] );

proposals/026-initialization-in-slang-2026.md
tangent-vector: I suspect that all of us in the conversation are already aware of it, but I want to make sure that anybody reading this is aware that the person most directly responsible for the pervasive acceptance of null pointers refers to it a ["billion-dollar mistake"](https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare/).

To be clear, Slang has never been as zealous about these kinds of issues as something like Rust, and it is likely we never *will* be quite as good about memory safety as Rust. That is by intention/design, rather than just because of a lack of awareness on our part as language designers and compiler writers.

The reality is that the kind of people who are writing code in Slang are the same kind of people who want to be able to write code that might dereference null without their compiler forcing them to fill out forms in triplicate noting that they are being A Very Naughty Programmer by doing so. So... I get it.

For me personally, as a language designer, the thing I care most about is the Path of Least Resistance.

If the kind of Slang code that is easiest to write happens to be correct, safe, and fast, then we are winning on all fronts. If there are cases where Slang makes it *easier* to write incorrect (or unsafe, or slow) code rather than some well-known correct/safe/fast alternative, we should look carefully at why that is.

The approach to initialization that Slang has been trending toward, and that C/C++ tend to use is *known* to be less correct/safe than the alternatives, and projects like Rust have shown that those alternatives can have comparable performance (although I'll concede that there are always outlier cases).

The whole point of the proposal I've written is that programmers who are happy enough with the C/C++ state of affairs can opt into it pretty easily (write `= default` everywhere you can, make all your types conform to `IDefaultable`, and wrap anything the compiler cares about in `Optional<T>` or `MaybeUninitialized<T>`), but simply *not* doing those things takes less effort, and leads to a programming model where the compiler is assisting the programmer by helping them see where they might have forgotten to initialize a variable, or neglected to think about the value that they wanted to put in some struct field.

tangent-vector:
> ...
> For example, something that achieves the same goal as this (albeit not necessarily written the same way):
> 
> ```
> Wave w;
> w.amplitude = 100.0;
> w.phase = 0.0;
> ```
> 
> Is still essential in my view. Unless we can absolutely guarantee that the compiler will always eliminate any redundant initialization, we need to retain the ability to bypass it. We definitely don’t want to end up in a situation where HLSL can produce faster code than Slang simply because we’ve tied the user’s hands on this.
> 

I believe that for 99% of use cases that want to do this sort of thing, they programmer would just set up their types to have a trivial `= default` and then write:

```
Wave w = default;
w.amplitude = 100.0;
w.phase = 0.0;
```

That doesn't give you any kind of 100% rock-solid guarantee that the compiler will eliminate the redundancy, but in any case where the compiler can determine that the defaulted value for, e.g., `w.phase` will be overwritten, it will make the logic that computed that value dead, and it would be eliminated unless there is a reason that the compiler cannot determine it is side-effect-free (which would seemingly only occur if we are doing full separate compilation and the function that computes the initial value is in another module and wasn't marked as being side-effect-free).

As I've said up-thread, the intention with this proposal is to start with The Simplest Thing That Could Possibly Work, so that all of us in the conversation can see what things might look like with all the special cases eliminated.
One of the most obvious special cases we might decide to add back in is the ability to write code just like your original:

```
Wave w;
w.amplitude = 100.0;
w.phase = 0.0;
```

in cases where `Wave` is determined to be a sufficiently "C-like" type. The semantics in that case would seemingly have to be that `w` would be fully *uninitialized* at the point of declaration, and would only become initialized one all of its fields have been explicitly written to.

Note that as soon as `Wave` has a user-defined constructor *or* any of the fields of `Wave` defines a default-value expression, then the `Wave` type doesn't really qualify as "C-like" and we are right back in the world where it doesn't make sense to do `w.amplitude = ...` until *after* somebody has run a constructor to initialize `w`. And once you need to construct `w`, the difference between using `= default` to do it vs. having the compiler auto-magically call a constructor behind your back (for some variables but not others...) is pretty much just a syntactic one, and your concern about whether the compiler can eliminate the double-initialization applies equally to this proposal and to C++ style pervasive automagical default constructors to "initialize" things.

I hope its clear that enough that in the case of a user-defined constructor on `Wave` the compiler really has no alternative but to insist that a call to at least one constructor happens before anybody uses `w`.

In the case where `Wave` doesn't have any user-defined constructors but *does* set default initial values for some of its fields, it may be harder to see why there's a problem... but trying to actually work out the details makes you quickly realize that its a tar pit.

The only reasonable way out is to *either* give every field in `Wave` an `= default` and then do `Wave w = default` and trust the compiler to eliminate the redundancy (since all your default values were simple and had no side effects), *or* to have a constructor for `Wave` that leaves it in a partially-initialized state.
The easiest way to express the latter condition is to wrap the specific fields of `Wave` that you want to support being in an uninitialized state with `MaybeUninitialized<T>`.

> Aside: if we can push through more stringent checking for initialization in the compiler front-end, then it would be a lot easier to change the logic in the back-end that currently writes zero values to initialize variables that it cannot prove are initialized (to avoid having compilers like dxc yell at us...) and have it instead write NaNs or `0xDEADBEEF` or similar intentionally-conspicuous values.

tangent-vector:
> With this:
> 
> ```
> int a[10];
> for(int i = 0; i < 10; i++)
>     a[i] = someFunc(i); // ERROR: `a` is being accessed while uninitialized
> ```
> 
> I’d really prefer not to have to rewrite that using unfamiliar or verbose syntax — and I suspect most users would feel the same. One of Slang’s big strengths is being easy to pick up and use; that kind of change risks undermining that.

Strong agree. Despite what I've written in the proposal, in my gut I think there's no way out of this particular issue.

I suspect that even in the 2026 language version, Slang will need to support a certain amount of field-by-field or element-by-element initialization logic, and the question is really around what we can support without breaking things too badly.

tangent-vector:
> A potential concern (though might be my lack of understanding). If we prevent function calls inside constructors, it becomes difficult to share initialization logic across multiple constructors. It would be unfortunate if we had to choose between duplicating code or initializing everything by default and then overwriting it — both approaches come with drawbacks in terms of clarity and performance.

To make sure we are on the same page:

* During part (1) of the proposed flow for constructors, you cannot make calls to member functions of `this`, but calls to other ordinary functions are allowed. This is comparable to the way that many C++ compilers will issue a strong warning if they see you doing member function calls on `this` during your base/member initialization list (because the order of initialization might not match what you think...).

* During part (3) of the proposed flow, you can make arbitrary member function calls on `this`, and the semantics are well defined and follow the Principle of Least Surprise, unlike the situation in C++ where calls to member functions *anywhere* during the body of a constructor can have unpredictable results. With this proposal Slang simply restricts member function calls to the situations where they can actually be given well-defined and unsurprising semantics.

To reiterate something I said up-thread: the steps (1), (2), and (3) that I describe for Slang constructors mirror how C++ does things, just with different syntax. C++ does step (2) via the base initializers, then it does (1) via member initializers, and then it does (3) by running the body of the user-defined constructor.

The main things that differ from C++ in this proposal are:

* You actually get a guarantee that your entire object/value is in a usable state for (3), so things like member function calls in the constructor will actually mean something for Slang `class`es, compared to where they are a known dangerous thing with C++ constuctors.

* In order to enable those guarantees, we swap the order of (2) and (1), which typically won't matter except in types that were too clever for their own good anyway. No matter what, a derived type gets the opportunity to run code both *before* and *after* the constructor the base type, which is strictly more powerful than being restricted to only one or the other.

* We don't allow arbitrary access to the contents of `this` (including member function calls) during (1) which, again, is something that C++ compilers strongly discourage even when the language allows it, and also isn't something the compiler can ever be confident isn't an error. If you really need to call arbitrary code, wait until (3), and just stick whatever default values you need to into the fields of your type first.

proposals/026-initialization-in-slang-2026.md
```diff
@@ -0,0 +1,444 @@
...
+
+1. Zero or more statements that do not access `this` or any of its members.
+
+2. A call `__init(a, b, c, ...);` to another constructor of the same type.
+
+3. Zero or mor statements that are allowed to access `this` as a fully-initialized value
+
+The body of a designated constructor must conform to the following sequence of operations:
+
+1. Zero or more statements that may assign to fields of the current type through `this`, but may not otherwise access `this` (notably: cannot access derived fields).
```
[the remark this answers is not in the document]
tdavidovicNV: aaah, field derived from the parent, got it.

I find the fact that the derived type must be initialized before the parent type extremely counterintuitive... in just about everything I do, parent has to be valid before child is.

proposals/026-initialization-in-slang-2026.md
```diff
@@ -0,0 +1,444 @@
...
+
+    int a[10];
+    for(int i = 0; i < 10; i++)
+        a[i] = someFunc(i); // ERROR: `a` is being accessed while uninitialized
+
+This proposal may live or die by the issue of supporting arrays cleanly.
+
+First, it is important to note that in cases where the element type of the array can be default-constructed, an array can be initialized with `default`:
+
+    Thing t[10] = default; // works if `Thing()` is valid
```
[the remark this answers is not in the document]
juliusikkala: Encoding uninitialized-ness in the type system worries me a bit. How would the below situation behave?

```slang
// Expects the array to be initialized
void f(inout Thing arr[4])
{
    ...
}

void main()
{
    MaybeUninitialized<Thing> arr[4] = default;
    for (int i = 0; i < 4; ++i)
        arr[i] = Thing(i);
    f(arr);
}
```

Should `f` take `MaybeUninitialized<Thing>` instead? If so, can it then be called with a regular `Thing[4]` as well? I guess the `BoundedSizeDynamicArray` may help here, but it seems like that will likely "waste" a register (unless there's a good optimizer) which I'm not very keen on since high register pressure is already a fairly common performance issue for me.

proposals/026-initialization-in-slang-2026.md
```diff
@@ -0,0 +1,444 @@
...
+
+#### Synthesized Constructor
+
+If a user-defined type doesn't declare any explicit `__init` declarations directly in its body, it is a candidate for having a constructor synthesized.
+The synthesized constructor gets a parameter for each field, with the corresponding type and name.
+Each parameter is optional only if the corresponding field has a default value expression.
+
+The synthesized constructor is a designated constructor.
+It initializes each direct field of the type in declaration order, to either the value of the coresponding parameter (if an argument was provided), or using the default-value expression for that field (if no argument was provided).
+Then it invokes the constructor of the concrete base type (if there is one) with zero arguments.
```
kaizhangNV, line 282: I'm not sure how this is going to work?
```
struct Base
{
    int a;
}
struct Derive : Base
{
    int b;
}
```

so the synthesized constructor of `Derive` will be like
```
__init(int b)
{
    this.b = b;
    super();
}
```
since `super()` is not existed, we would synthesize a constructor for type `Base` with signature as `__init(int a)`, so we seems cannot "invokes the constructor of the concrete base type (if there is one) with zero arguments."

Am I understanding this wrong?

PR #23, opened 2025-06-04
Conditional

csyonghe:


tangent-vector (approved):
Looks good. We’ve talked about the need for this type several times before, so I’m pleased to see a proposal for it.

tangent-vector:
The previous comment had me realize that this proposal should probably go into greater detail about the behavior of `Conditional<T,v>` when it comes to layout.

It is implicit in the goals of the proposal that when the condition is `false` the value consumes no space (and the compiler already handles zero-sized values well), and I suppose the question of what gets reflected for such a field is no different than any other generic-parameter-dependent type.

However, given that a common use case for this feature would be vertex attributes, there is potentially value in assigning varying input/output locations to `Conditional` fields *as if* they always contain a value. In principle, doing so would mean that, for example, a vertex shader specialized to output position, normal, and texcoord could be used with a fragment shader specialized to receive only position and texcoord (since the layout for the texcoord field wouldn’t be affected by the presence/absence of the normal field).

I’m not sure if there’s enough value to be gained from that behavior to make it worth implementing, but I do think the proposal should be clear about what we intend to implement.

proposals/029-conditional.md
```diff
@@ -0,0 +1,179 @@
...
+
+```slang
+[shader("vertex")]
+Vertex<Normal, Color> vertMain<Normal : IConditional<float3>, Color : IConditional<float3>>(VertexIn inputVertex)
+{
+    ...
+}
+```
+
+And `vertMain` can then be specialized with `vertMain<Real<float3>, No<float3>>`. This can achieve the same goals, but it is
```
kaizhangNV, line 176: Yea, I definitely agree this helper is more cumbersome than the `Conditional<T, bool>` syntax.

LGTM

kaizhangNV (approved):

kaizhangNV (commented):
With respecting to the reflection.

If user query the variable which is Condition<T, false>, what will happen?
We just return `nullptr`?

csyonghe:
The size for Conditional<T, false> should be 0.

csyonghe:
Added discussion of layout and metal's function_constant.

csyonghe:
@kaizhangNV A `Conditional<T, false> value;` field will still be reflected as a variable/field whose type is `Conditional<T, false>` and whose size is 0 in our reflection API.

PR #24, opened 2025-06-12
#30 - Coherent Pointers & Pointer Access Proposal

ArielG-NV:
Resolves: https://github.com/shader-slang/slang/issues/7408

This proposal is for implementing (only for SPIR-V targets) coherent loads and stores through pointers. This proposal also implements ReadOnly (immutable) pointers & ReadWrite pointers.

A Slang playground example is provided [here](https://shader-slang.org/slang-playground/?target=SPIRV&code=eJzNVlFP2zAQfs-vOCE0pUCBjm0PlFXa2KZNW0cFFTxEkeu2TrGa2pnthFUb_31nO01Dga0gNq2qqvju8919313sEjJhgik-OurvQMoM6JHMGBxCl82kmp_ZVSeIFPuWc8XCs945-fzxlBR5OqWCzByIzOSYpY040EblIwPH8pIpJkzPqOBHAPjpbwHJjGoHgVtGH6QasU8i5YLFzkIIF9yEiENYw5n8TudEG7wGF8Cur8swe3twUjCVSjo-pWIKXEOzBVqCuaQGBuy7YUJzKQYawpRPGSRSwYD00ZErNgAqxjA4ljLrUjNoLEJe0oJBprhU3MxBYgKMx2DMEpqnxm6o2A0gycXIYAqaInjXk6vXFDZbDc-wD9YW5lwYkEmimVmlWVBVenqOr6O9XZraFUwxrF6AzrgqCNWzyrGMZD8n2fsFf9j4Tds22ivbjmlGh9zygXMH95PQtWCMsQr_gqxgc7OPdWmrz-aSwpuUTwQbw0_o0inrSaTO1DnXfJgyjIT2r1L0FC-oWXituQWbbgiXma5v9v1ehbWROKM1iXdwsBRL0CXY1TlNc_Z3RX-s6g-V_R_pvio8ITof6pHimbkxyM1Of0XWaJYbariYxJUJgfeo5tu26FjVqVr_q6fJvUHKDrmXrCxrNcB1YIkEAXnKMy8iJBeaJqx2pr2nKp3HQe2ksJlclg7gj6LIeDu8219Y8juA8r56QZZnhWdrx9Wfhg62e8e4lkLcGTy0p2vbyfB_iNB8MhGajxFhNMyTBA_4Yym0ocLoMgPm3YIJknzr_AjGq-v04szdb3h3jL39CHFR3IFhiboHgnRzk-VY2rp6FpKP7R0kTJdyETbqZXnm9eKsq84UYTv1Nh4eXkg1nSiZZx27nVwtlhho3Y3hMmNjnZTvWMFHzOcbu-c_Jiu33MzkT54bVUf7McbymkatuH0b4f8t1Nbb0PKwctfzeBWCQdcJ1FwJdHA70EFcqzpquWJb-25XlaBSpLbYhgMfesk_qm-uYX3aqoR28LAZoErRub_v1m7_s-GiIvyWE7CI48t8ad-pX0EvRsM) to test how `MakePointerVisibleKHR` and related operands function. This is not an implementation of the proposal.

### DIFF
--- a/proposals/030-coherent-pointers.md
+++ b/proposals/030-coherent-pointers.md
@@ -63,7 +63,7 @@ public enum CoherentScope
     Subgroup = MemoryScope::Subgroup,
     Invocation = MemoryScope::Invocation,
     QueueFamily = MemoryScope::QueueFamily,
-    ShaderCallKHR = MemoryScope::ShaderCallKHR,
+    ShaderCall = MemoryScope::ShaderCallKHR,
     //...
 }
 


proposals/030-coherent-pointers.md
```diff
@@ -0,0 +1,190 @@
...
+
+`CoopVec` and `CoopMat` load data into their respective data-structures from other objects using `CoopVec::Load`, `CoopVec::Store`, `CoopMat::Load`, and `CoopMat::Store`. Due to this design, we will add coherent operations to `CoopVec` and `CoopMat` by modifying `CoopVec::Load`, `CoopVec::Store`, `CoopMat::Load`, and `CoopMat::Store` to complete coherent operations if given a `Ptr` with the flag `PointerFlags.Coherent` as a parameter. Syntax required to use the methods will not change.
+
+### Support Casting Pointers With Different Coherent `MemoryScope`
+
+We will allow pointers with different coherent `MemoryScope`’s to freely cast between each other. For example, `Ptr<T, PointerFlags.Coherent, MemoryScope.Device>` will be castable to `Ptr<T, PointerFlags.Coherent, MemoryScope.Workgroup>`.
+
+### Support The `globallycoherent` Keyword
+
+Slang will implement support for use of the `globallycoherent T*` keyword. This will be an alias to `Ptr<T, PointerFlags.Coherent, MemoryScope.Device>`.
```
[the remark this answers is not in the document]
csyonghe: This syntax is complex enough that we may want to completely disallow it in the initial implementation with an error on things like `globallycoherent int*`, and then support it in a separate languauge proposal.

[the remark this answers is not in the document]
ArielG-NV: `CoherentPtr<T, MemoryScope>` worries me a bit since down the line there may be situations where a pointer abides to 2 different memory scopes for whatever reason SPIR-V requires.

Maybe the eventual solution here is named generic parameters: `Ptr<type=T, flags=PointerFlags.Coherent, scope=MemoryScope.Device> ptr;`

proposals/030-coherent-pointers.md
```diff
@@ -0,0 +1,190 @@
...
+### Frontend For Coherent Pointers
+
+We propose to implement coherence on a per-operation level for only SPIR-V targets. This will be accomplished through modifying `Ptr` to include new generic arguments `PointerFlags flags` and `MemoryScope coherentMemoryScope`.
+
+```c#
+
+public enum PointerFlags
+{
+    None = 0,
+    Coherent = 0b1
```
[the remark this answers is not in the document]
tangent-vector: I would worry about overloading a single flags enum to handle too many different axes here. The following all seem, to me, to be reasonably orthogonal (and thus merit distinct generic arguments)

* The *type* of the storage location being pointed to

* The *address space* where the storage location resides

* The *access* modes that are allowed to be performed through the pointer (read-only, write-only, or read-write)

* The *coherence scope* for operations performed through the pointer.

proposals/030-coherent-pointers.md
```diff
@@ -0,0 +1,208 @@
...
+}
+
+__generic<T, uint64_t addrSpace=AddressSpace::UserPointer, PointerFlags flags=PointerFlags.None, MemoryScope coherentMemoryScope=MemoryScope.Device>
```
tangent-vector, line 60: It would be nice to fold the `PointerFlags.Coherent` part together with the `coherentMemoryScope` part, because there's a strong relationship between the two:

* It is meaningless to specify a scope without also specifying the `Coherent` flag, but this is allowed and means that code could look like it does something it doesn't. E.g., if they write `Ptr<T, PointerFlags.None, MemoryScope.WorkGroup>` that might *look* like there's something meaningful being communicated via the `WorkGroup` argument, but there really isn't.

* A programmer really shouldn't be saying something is `Coherent` without taking a second to think about the scope of coherency they want.

I would propose that the simplest fix is to support a "scope" that represents operations only being guaranteed coherent within the current thread/work-item/lane/whatever-you-like-to-call-it.

proposals/030-coherent-pointers.md
```diff
@@ -0,0 +1,208 @@
...
+
+coherent T groupshared* // a pointer to groupshared coherent memory: Ptr<T, AddressSpace::GroupShared, PointerFlags.Coherent, MemoryScope::WorkGroup>
+groupshared coherent T* // same as above
+coherent groupshared T* // same as above
+...
+int* groupshared // a groupshared variable that is a global pointer: groupshared Ptr<T, MemoryScope::Device>
+int* coherent groupshared* // a pointer to coherent groupshared memory: Ptr<T, AddressSpace::GroupShared, PointerFlags.Coherent, MemoryScope::WorkGroup>
+int* coherent* groupshared // a groupshared variable that is a pointer to coherent global memory: groupshared Ptr<T, PointerFlags.Coherent, MemoryScope::Device>
+
+globallycoherent T* // pointer which is coherent to global memory: groupshared Ptr<T, PointerFlags.Coherent, MemoryScope::Device>
```
tangent-vector, line 141: I strongly object to a lot of this, just because it adds more modifiers and more keyword soup to Slang. We should be very cautious about adding tons of keywords and *especially* about keywords that can be used to qualify pointers...

proposals/030-coherent-pointers.md
```diff
@@ -0,0 +1,208 @@
...
+
+If `PointerFlags.Coherent` is present, all accesses to memory through this pointer will be considered coherent. The coherent operation will be coherent to the memory scope `coherentMemoryScope`.
+
+### Support For Coherent Buffers and Textures
+
+Any access through a coherent-pointer to a buffer/texture is coherent.
+
+```c#
+RWStructuredBuffer<int> val; // Texture works as well.   
+Ptr<int, PointerFlags.Coherent, MemoryScope.Subgroup> p = &val[0];
```
tangent-vector, line 75: This line of code already opens up a line of questions that I think is really important for this proposal to address: what are the conversion rules for `Ptr` values under this proposal? Can a pointer with `Coherent` be converted to one without? Vice-versa? Can one with a narrow coherency scope be converted to one with a wider scope? Why or why not?

tangent-vector (commented):
I think this is important stuff to expose, but it also touches on a lot of big-picture topics in the language design that I wouldn't want us to bake too soon.

proposals/030-coherent-pointers.md
ArielG-NV: This will be implemented (per param, 1 unique enum) as per suggested in https://github.com/shader-slang/spec/pull/24#discussion_r2155469776

example (not final):
```
enum CoherentPtrFlags
{
    NotCoherent,
    CoherentToDevice,
    CoherentToWorkgroup,
    ...
}
```

proposals/030-coherent-pointers.md
ArielG-NV: Final decision made: We are to remove this feature as a "next step". We should not have a stance on this currently.

proposals/030-coherent-pointers.md
ArielG-NV: I think it is a give and take here (rather than just all bad as with `const`).

* Users are likely to mix Coherent and Non-Coherent operations (expected behavior), hence casting freely will be useful. It is expected that a user will always need access to 2 variations of a `CoherentPtr`
* This is a contrast to `const` since for all `const` variables, casting away `const` is always a bad idea and almost always a source of bugs. 
* Coherence is much more comparable to atomic operations, "do we want the types to be atomic? Do we want per operation atomics?".

proposals/030-coherent-pointers.md
```diff
@@ -0,0 +1,203 @@
...
+...
+let a = *p; // should be aligned load;
+let b = p.member; // should be aligned load, with alignment derived from both `MyType` and `member`'s type.
+```
+
+When loading data from a pointer `p` Slang will honor the alignment and emit an `OpLoad` with the SPIR-V `Aligned` memory operand, providing the argument `ALIGNMENT`. This will function alongside `coherent` pointers.
+
+### Additional Pointer Arguments
+
+`Volatile` and `Const` are planned features for `Ptr`.
```
csyonghe, line 137: Why `Volatile` when there is already `coherent`?

proposals/030-coherent-pointers.md
```diff
@@ -0,0 +1,203 @@
...
+}
+```
+
+If `coherentScope` is not `CoherentScope.NotCoherent`, all accesses to memory through this pointer will be considered coherent to the specified memory scope (example: `CoherentScope.Device` is coherent to the memory scope of `Device`).
+
+We will also provide a type alias for user-convenience.
+
+```c#
+__generic<T, CoherentScope coherentScope>
+typealias CoherentPtr = Ptr<T, AddressSpace::UserPointer, coherentScope>;
```
csyonghe, line 79: I think this should be:
```
typealias CoherentPtr<T> = Ptr<T, AddressSpace::UserPointer, MemoryScope::CrossDevice>;
```

For anything else, the user should just use `Ptr<>` type.

proposals/030-coherent-pointers.md
```diff
@@ -0,0 +1,203 @@
...
+
+Any access through a coherent-pointer to a `groupshared` object is coherent; Since Slang does not currently support pointers to `groupshared` memory, this proposal will extend the existing `AddressSpace::GroupShared` implementation for pointers as needed.
+
+### Support For Coherent Cooperative Matrix & Cooperative Vector
+
+`CoopVec` and `CoopMat` load data into their respective data-structures from other objects using `CoopVec::Load`, `CoopVec::Store`, `CoopMat::Load`, and `CoopMat::Store`. Due to this design, we will add coherent operations to `CoopVec` and `CoopMat` by modifying `CoopVec::Load`, `CoopVec::Store`, `CoopMat::Load`, and `CoopMat::Store` to complete coherent operations if given a `CoherentPtr` as a parameter. Syntax required to use the method(s) will not change.
+
+### Support Casting Pointers With Different `CoherentScope`
+
+We will allow pointers with different `CoherentPtr` to be explicitly castable to each other. For example, `CoherentPtr<int, CoherentScope.Device>` will be castable to `CoherentPtr<int, MemoryScope.Workgroup>`.
```
csyonghe, line 105: ... with different `CoherentScope` to be explicitly...

proposals/030-coherent-pointers.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Order of Implementation
+
+1. Support for workgroup memory pointers.  
+2. Frontend for coherent pointers  
+3. Support for coherent buffers and textures  
+4. Support for coherent workgroup memory  
+5. Support for coherent cooperative matrix & cooperative vector  
+6. Support casting pointers with different coherent `MemoryScope`  
+7. Support the `globallycoherent` keyword
```
csyonghe, line 115: Probably delete 7.

csyonghe: We should make `globallycoherent T*` an error.

csyonghe (commented):
Looks good overall, some minor comments.

csyonghe:
Perhaps we should just extend this proposal to also introduce `access`.

proposals/030-coherent-pointers.md
ArielG-NV: `Volatile` is functionally different than `Coherent`: "This access cannot be eliminated, duplicated, or combined with other accesses.".

Therefore `Volatile` is a potential future addition (although I doubt high priority).

proposals/030-coherent-pointers.md
```diff
@@ -0,0 +1,225 @@
...
+public enum CoherentScope
+{
+    NotCoherent = 0xFF,
+    CrossDevice = MemoryScope::CrossDevice,
+    Device = MemoryScope::Device,
+    Workgroup = MemoryScope::Workgroup,
+    Subgroup = MemoryScope::Subgroup,
+    Invocation = MemoryScope::Invocation,
+    QueueFamily = MemoryScope::QueueFamily,
+    ShaderCallKHR = MemoryScope::ShaderCallKHR,
```
csyonghe, line 66: Remove "KHR".

ArielG-NV: removed 👍

PR #25, opened 2025-06-13
Optional constraints proposal

### LINKED ISSUE #18: Optional witnesses
As briefly discussed in https://github.com/shader-slang/slang/issues/6885 (filing an issue here since I intend to eventually implement this and this needs a spec).

These would allow a generic type parameter to optionally conform to some interface. This mechanism also needs to provide some method to check this conformance for a block of code and allow using the related interface in that block.

Copying from Yong He's reply to that issue, one possible syntax for this would be:
```slang
// `w` refers to the optional `T:IThing` conformance.
// Internally, `w`'s type is `Optional<SubtypeWitness<T,IThing>>`.
void tryF<T : IThing? as w>(inout T val)
{
    if (w)
    {
       // here the compiler can assume `T:IThing`.
        val.f();  
    }
}
```

While this is fine, I'd like to bring up an alternative syntax for discussion as well:
```slang
void tryF<T : IThing?>(inout T val)
{
    do
    {
       // here the compiler can assume `T:IThing`.
        val.f();  
    }
    where T: IThing; // this should fail to compile if T doesn't have an optional conformance to IThing
}
```

This syntax:
 \+ Prevents having to deal with cases like `if(w || somethingElse)` or `if(w1 && w2)`
 \+ Keeps the type declaration in the generic parameter simpler, as it doesn't need to introduce a new variable
 \- Likely always causes user to write more code

Also, is there any plan to add `?` syntax for optional types, like `Type? == Optional<Type>` in Swift? If so, this would clash with that, since `T : IThing?` would be ambiguous in that it could either refer to an optional conformance or a type that derives from `Optional<IThing>` (although, now that I think about this, the latter case seems impossible to occur in legitimate code, so maybe this is OK). Optional conformances could perhaps be denoted with `T ?: IThing` if this is a real problem.

I'm fine with writing the proposal myself simultaneously to implementing the feature, once there's some rough plan / consensus on how it should work.

juliusikkala:
Closes #18. Based on https://github.com/shader-slang/slang/issues/7201#issuecomment-2902811333. Implementation ongoing in https://github.com/shader-slang/slang/pull/7422.

@csyonghe do I also add you as a co-author? The syntax is entirely based on your comments.

### DIFF
--- a/proposals/000-optional-constraints.md
+++ b/proposals/000-optional-constraints.md
@@ -65,11 +65,11 @@ Then, `MyTrait` can be unconditionally required and the optional functionality j
 Proposed Approach
 -----------------
 
-A new `where optionally` syntax is introduced:
+A new `where optional` syntax is introduced:
 
 ```slang
 void release<T>(T t)
-    where optionally T : IThing
+    where optional T : IThing
 {
     t.doThing(); // ERROR, it's possible that T doesn't implement IThing
     if (T is IThing)
--- a/proposals/000-optional-constraints.md
+++ b/proposals/000-optional-constraints.md
@@ -1,0 +1,114 @@
+SP #000: Optional constraints
+=================
+
+Extending [SP #001], we should allow optional `where` statements that can be statically checked for sections in a function body. This allows users to succinctly express optional behaviour depending on whether an interface is implemented or not.
+
+Status
+------
+
+Status: Design Review
+
+Implementation: [PR 7422](https://github.com/shader-slang/slang/pull/7422)
+
+Author: Julius Ikkala
+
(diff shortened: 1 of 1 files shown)


csyonghe:
I am open to supporting `T: maybe IThing` as a syntax if we find it to be convenient, but as you pointed out, optionality is an orthogonal concept to the kind of constraints, so we will need a general syntax for it anyways that will work for e.g. type equality constraint, value constraint(in the future) etc.

csyonghe (approved):
what's the implementation status of this? should give it a number, e.g. 039. Otherwise looks good to me.

PR #26, opened 2025-06-13
Add proposal for interface default implementations.

csyonghe:


proposals/030-interface-method-default-impl.md
```diff
@@ -0,0 +1,76 @@
...
+struct Impl2 : IFoo
+{
+    // OK, overriding the default implementation.
+    int getVal() { return 1; }
+}
+```
+
+Both static and non-static methods can have default implementations.
+
+A default implementation can be marked as `[Diffeerentiable]` for automatic differentiation.
```
kaizhangNV, line 49: typo `Diffeerentiable`

proposals/030-interface-method-default-impl.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+An interesting question is what should the expected behavior be when there are overlapping method declarations in interface hierarchies. Consider the following:
+
+```slang
+interface IBase
+{
+    int getVal() { return 0; }
+}
+interface IDerived
```
tangent-vector, line 81: I assume this should have `IBase : IDerived`.

tangent-vector (approved):
This looks good, and is enough to start with. I do think that there are going to be corner cases and semantic wrinkles that will come up in implementation.

One basic question is whether we should force users to mark default implementations in an interface as `virtual` to make it clear that they are still overridable requirements and not the same as, say, extension methods.

The proposal should make clear what kinds of interface requirements it does and does not apply to. My guess is that we don’t want to allow defaults for associated types at this point.

We should consider whether we should diagnose cases where a derived type has a method that matches on name but isn’t considered and override for a base-type method. The C# solution of requiring either `override` or `new` in these cases is nice. This is one of those cases where allowing something without a diagnostic in the near term may make it hard to add more constraints in the future.

A related note on implementation is that we will likely need to change the approach to finding witnesses for interface requirements so that it considers a less-good signature match in a derived type before an exact match in a base type (including the original interface).

csyonghe:
After talking to users, the general preference is to avoid the `virtual` keyword since it causes confusion on whether that means dynamic dispatch and virtual function tables. By default, we want to treat everything inside an interface as a requirement, and if we want to treat things differently, perhaps it needs to be an explicit opt-out, such as `nonoverridable` `nonrequirement` etc.

Updated the spec to require explicit `override` on the conforming type side, if the corresponding interface requirement has a body.

Also clarified that this proposal is only about method requirements, not about properties, subscripts, associated types/constants.

tangent-vector (approved):
Looks good.

I think the user feedback is completely reasonable, and I'm swayed by it. That choice effectively means that declarations in an `interface` are implicitly `abstract` if they have no definition, and implicitly `virtual` if they have a definition. We should eventually allow the declarations to be explicitly marked with those keywords (or whatever the equivalents end up being, one we need and add them), but it's reasonable to never require them in an `interface` body, where they are kind of implicit.

The one thing I'd like to confirm, if we go down that road, is that we should *not* allow any declarations inside of an `interface` that do not represent requirements. That would mean being explicit that things like `typealias`es, `struct`s, and `enum`s are not allowed in the body of an `interface`. Or, if we *do* want to allow those things in an `interface` body, I would propose that they should have to be marked with `final` or `sealed` or whatever our keyword for "this cannot be overridden" will end up being.

PR #31, opened 2025-07-31
A proposal to deprecate APIs we never intended

tangent-vector:
There's a lot of cruft in `slang.h` and the `slangc` command-line options that was never really meant to be used, and a lot of other cruft that we knew was a liability from the moment it landed. This is a proposal to mark as much stuff deprecated as currently seems possible, with the hope that over time we can clean up the implementation by not needing to support too many legacy features and in-retrospect-poor design choices.

### DIFF
--- a/proposals/032-deprecate-unintended-apis.md
+++ b/proposals/032-deprecate-unintended-apis.md
@@ -1,0 +1,219 @@
+SP #032: Deprecate Public APIs That Shouldn't Exist
+===================================================
+
+We propose to deprecate a large number of functions, enumerants, etc. from the Slang public API, and command-line options for `slangc`, in cases where any of the following is true:
+
+* the features were never intended for public use
+
+* the features represent design choices we regret
+
+Status
+------
+
+Status: Implementation In-Progress
+
+Implementation: -
+
+Author: Theresa Foley (@tangent-vector)
+
+Reviewer: -
+
+Background
+----------
+
+This proposal touches on two main categories of features of the Slang compiler codebase:
+
+### Features That Were Never Intended for Public Use
+
+Slang's build process involves some "boostrapping" steps, including needing to invoke the Slang compiler on the source code for the builtin modules, so that the compiled binary for the core module can be embedded into the Slang compiler itself.
+Unfortunately, at the time that the bootstrapping process was implemented, the hooks that were required to implement it got declared in the public `slang.h` header, thus enabling others to use those hooks, even if they were never intended to be part of the supported API of the compiler. The same applies to several command-line options for `slangc` (which also surface as enumerants in `slang.h`) that are only intended to be used during the bootstrapping process.
+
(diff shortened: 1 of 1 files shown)


proposals/032-deprecate-unintended-apis.md
```diff
@@ -0,0 +1,221 @@
...
+* option `-report-downstream-time` should be marked as internal
+
+* option `-report-perf-benchmark` should be marked as internal
+
+* option `-report-checkpoint-intermediates` should be marked as internal
+
+* option `-source-embed-style` should be deprecated
+
+* option `-source-embed-name` should be deprecated
+
```
juliusikkala, line 172: I am actively using these options, they're useful for small applications that want to embed SPIR-V in the executable so that they don't need to care about a filesystem. They also mirror equivalent `glslc` / `glslang` functionality, making `slangc` a drop-in replacement in an existing CMake shader compilation setup.

I have no issues with the other deprecations listed.

juliusikkala: I guess I'm also not the only person using this functionality, here's someone recently on Discord asking how to do exactly what these options do: https://discord.com/channels/1303735196696445038/1397550571321299006

swoods-nv:
Tagging @danginsburg for input and awareness as well

proposals/032-deprecate-unintended-apis.md
```diff
@@ -0,0 +1,221 @@
...
+* option `-unscoped-enum` should be marked as undesirable (and subject to future deprecation/removal)
+
+* option `-preserve-params` should be marked as undesirable (and subject to future deprecation/removal)
+
+* option `-reflection-json` should be marked as undesirable (and subject to future deprecation/removal)
+
+* option `-msvc-style-bitfield-packing` should be marked as undesirable (and subject to future deprecation/removal)
+
+* option `-fspv-reflect` should be marked as undesirable (and subject to future deprecation/removal)
```
natevm, line 183: Isn’t this required to support Khronos’s SPIRV reflection library? https://github.com/KhronosGroup/SPIRV-Reflect if I understand right, then IIRC Valve was a proponent that we support SPIRV-Reflect decorators.

 If this is true, then I’d probably object to removing this command line argument.

proposals/032-deprecate-unintended-apis.md
```diff
@@ -0,0 +1,221 @@
...
+Some things got put into the `slang.h` API that were over-designed from the start, and never saw enough uptake among developers to justify the cost of supporting all the bells and whistles.
+Examples include:
+
+* `ISlangCastable` and `ISlangClonable`
+
+* `ISlangFileSystemExt`
+
+* `ISlangWriter`
+
+* The ability for users to override the "prelude" code for particular target languages and/or downstream compilers
```
CrossVR, line 45: We are currently using a custom build of Slang that uses our own C++ prelude in addition to some other changes. This feature could be a potentially useful way for us to use the official build of Slang.

Our intention is to contribute our changes back, but some of our prelude changes rely on LLVM compiler plugins so I'm not sure if all our prelude changes will make it in.

PR #33, opened 2025-09-09
Amendments to "017-shader-record" 

natevm:
Main change is to have the proposal language use `Type<T>` rather than `[[attribute]]` (the later could either be syntax sugar, or could result in a diagnostic)

Trimmed down some of the text regarding rates to help keep discussion focused.

### DIFF
--- a/proposals/017-shader-record.md
+++ b/proposals/017-shader-record.md
@@ -1,29 +1,40 @@
-SP #017: Disambiguate `uniform` via `[[push_constant]]` and `[[shader_record]]` 
-=======================================================================
+SP #017: Disambiguate `uniform` via `Payload<T>`, `HitAttribute<T>`, and `ShaderRecord<T>`
+============================================================================================
 
 This document proposes a first step toward resolving ambiguities and bugs that
 arise from the current usage of `uniform` (and to a lesser extent, `varying`)
 in Slang, particularly in ray tracing pipelines.
 
-<!-- It suggests introducing new types such as `PushConstant<T>` and
-`ShaderRecord<T>`, which can be used to unify compile-time mappings across APIs
-like Direct3D and Vulkan. These new types act as clearer, more consistent
-markers of data binding points, compared to the over-loaded semantics
-of `uniform` parameters today. -->
+This document proposes two concrete shifts to remove ambiguity around `uniform` and ray-tracing
+entry point parameters:
 
-It suggests promoting Vulkan's `[[push_constant]] uniform T` and 
+- It proposes to make `uniform` universally mean "push constants" at entry points across all backends 
+  and pipeline stages with Slang version 2026. Where push constants are not first-class, Slang maps 
+  them to the closest equivalent (e.g., D3D root constants or a small constant buffer; OptiX launch parameters).
+- It suggests introducing new types such as `PushConstant<T>`, `ShaderRecord<T>`, and `HitAttribute<T>`, 
+  which can be used to unify compile-time mappings across APIs like Direct3D and Vulkan. These new types act 
+  as clearer, more consistent markers of data binding points, compared to the over-loaded semantics
+  of `uniform` parameters today.
+
+Together, these changes provide clear, explicit markers of data binding points without overloading
+`uniform`, `in`, `out`, and `inout` semantics on ray tracing entry points.
+
+<!-- Previously, we were considering [[annotations]], but with some discussion, Type<T> would enable future
+language improvements like pipeline interfaces. -->
+
+<!-- It suggests promoting Vulkan's `[[push_constant]] uniform T` and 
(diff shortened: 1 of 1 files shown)


proposals/017-shader-record.md
```diff
@@ -249,20 +284,29 @@ Proposed Approach
...
-
-2. **Entry-Point Parameter Rules**
+    pressing distinction. -->
+
+1. **Entry-Point Parameter Rules**
+    * Parameters declared `uniform` at entry points always map to push-constant-like storage
+      (push constants, root constants, or a small constant buffer) regardless of stage/backend.
+    * Parameters typed as `ShaderRecord<T>` map to per-shader-record data (Vulkan/OptiX SBT;
+      D3D local root signature).
+    * Parameters typed as `Payload<T>` map to payload registers. `in`, `out`, and `inout` apply,
```
csyonghe, line 305: My understanding of these wrapper types is that they should be treated as if they are pointer types. So `Payload<T>` is effectively a `T*`. It is not meaningful to have `inout T*` or `out T*` because that just means the pointer itself is mutable. Instead, we probably want to have `RWPayload<T>`, `WPayload<T>` to represent payload pointers with read/write or write-only access.

proposals/017-shader-record.md
```diff
@@ -275,20 +319,31 @@ Detailed Explanation
...
+
 4. **Migration Strategy**
+    * Existing Slang code using `[[vk::push_constant]] ConstantBuffer<T>` remains valid but is
+      encouraged to migrate to `uniform` entry point parameters for clarity and portability.
+    * Existing code using `[[vk::shader_record]] ConstantBuffer<T>` remains valid; `ShaderRecord<T>`
+      entry point parameters are the recommended form. 
+    * If either `[[vk::shader_record]]` or `[[vk::push_constant]]` or `[[shader_record]]` or `[[push_constant]]`
+      appear on entrypoint parameters, these will become "syntax sugar" which is transformed by slang into
+      the appropriate wrapper type. `[[hit_attribute]]` might be considered for addition to allow for consistency.
+    * Legacy usage of bare (non-`uniform`) parameters in ray tracing entry points should result in a warning, telling
```
csyonghe, line 341: This should be an error after slang 2026.

proposals/017-shader-record.md
csyonghe: We should also just make it invalid to use `in` `inout` or `out` on ray tracing entrypoint parameters.

natevm: I’m not sure I’m aligned on the RWPayload though. With PAQs, users annotate individual fields within the structure passed as T as being read, write, or read write. It generally never makes sense to make the entire payload read-only. 

My thought was to instead look at the T being passed into the payload type and assume that that’s where PAQs information comes from. 

Then otherwise drop in/out semantics entirely

natevm:
Sorry, if I recall correctly, this proposal was never officially accepted. At the time, there was this open question about how to address the rates of entrypoint parameters. 

From there, the discussion evolved to autodifferentiability of parameters having different rates across different entry points which cannot easily be associated with one another unless a larger "pipeline" syntax was added to slang. 

I believe from there it was decided such a feature would be too large of a change to the language, but that this still did not fully justify the acceptance of this proposal. 

I might have the history there wrong. Might be worth revisiting if this proposal should be implemented or not.

jhelferty-nv:
Putting this on @skiminki-nv's radar.

proposals/017-shader-record.md
```diff
@@ -1,29 +1,40 @@
...
-`ShaderRecord<T>`, which can be used to unify compile-time mappings across APIs
-like Direct3D and Vulkan. These new types act as clearer, more consistent
-markers of data binding points, compared to the over-loaded semantics
-of `uniform` parameters today. -->
+This document proposes two concrete shifts to remove ambiguity around `uniform` and ray-tracing
+entry point parameters:
 
-It suggests promoting Vulkan's `[[push_constant]] uniform T` and 
+- It proposes to make `uniform` universally mean "push constants" at entry points across all backends 
+  and pipeline stages with Slang version 2026. Where push constants are not first-class, Slang maps
```
skiminki-nv, line 12: 2026 is feature complete. This should probably now target 202c. See https://github.com/shader-slang/slang/blob/c8ba6929456b8e9c1a097970d929f5bc003250f8/docs/user-guide/11-language-version.md

skiminki-nv (commented):
@natevm Were you planning to still update this PR?

skiminki-nv:
I think this PR is better handled by @tangent-vector or @csyonghe. See comment https://github.com/shader-slang/spec/pull/33#issuecomment-4774323760 for the open issues.

skiminki-nv:
Exchanging my assignment with Tess's.

PR #34, opened 2025-09-23
A Proposal for Shader Components

tangent-vector:
Looking for feedback.

### DIFF
--- a/proposals/032-shader-components.md
+++ b/proposals/032-shader-components.md
@@ -1,0 +1,1156 @@
+SP #032: Shader Components
+=================
+
+We propose to add a construct for authoring "components" in Slang.
+These components provide a unit of shader modularity and composition that sits between `struct` types and modules in granularity.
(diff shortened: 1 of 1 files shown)


juliusikkala (commented):
I quite like it, but I'm a bit worried how alien this feature may feel to developers. If the feature is not familiar and requires a long and complicated explanation, it may be difficult to get people to start using it.

natevm:
We spoke a lot about this already, but figured I’d chime in here too that I’m in favor of this feature proposal.

I tend to agree with @juliusikkala that the feature might go unused if it looks too "alien”, though I think this can be addressed. Like you said, main focus here is on the semantics.

This would address many of the somewhat pervasive issues I’ve had with slang projects over the years. 

These days I rarely have vertex or fragment shaders, but I _do_ have many compute kernels all in the same file. Eg a BVH builder can easily be 10 or so compute entrypoints in a single file. A “component” system like this will allow me to keep that code more “DRY” and maintainable.

PR #39, opened 2026-01-21
Pointer data layout parameter

### LINKED ISSUE #36: Data layouts for pointers
Slang currently allows buffer types to use a variety of data layouts, such as `ScalarDataLayout` or `Std430DataLayout`. However, this is not the case for pointers, which do not have an explicit control for what the data layout is. It is usually assumed to be the scalar layout.

I would like to see any of the following approaches to allow controlling data layouts with pointers:
1. Ability to annotate data layout of `T` with new generic parameter to `Ptr`: `Ptr<T, Access, AddressSpace, IBufferDataLayout>`
2. Ability to annotate data layout per-type, e.g.:
```slang
[DataLayout(ScalarDataLayout)]
struct SomeStruct {
    ...
}
```
3. Both the generic parameter and attribute, where the latter only defines a default layout.

For more context, see https://github.com/shader-slang/slang/pull/8960#discussion_r2593802720. This issue is blocking the implementation of https://github.com/shader-slang/slang/issues/9478.

Next, I'll try to provide some arguments for each approach.

## 1. New data layout parameter to `Ptr`

This is likely to be the approach requiring the least changes to the codebase, and shouldn't be a breaking change as long as a default value is provided.

It also allows a lot of flexibility; it would be possible to point to the same structure with different layouts, just as the data layouts work with buffers currently.

## 2. Struct attribute for layout

This fits the mental model of a structure itself defining the memory layout. This is likely a common mental model for many developers familiar with C or C++ and matches how the layout options are handled in C compilers (`[[gnu::packed]]`).

It would also make it possible to define the memory layout of stack-allocated variables in the LLVM emitter. The user can observe that layout, because pointers to stack variables can be created with the CPU targets.

This approach enables mixing and matching different memory layouts; e.g.
```slang
[DataLayout(Std430DataLayout)]
struct A
{
    float3 a;
    float b;
}

[DataLayout(CDataLayout)]
struct B
{
    bool a;
    bool b;
    bool c;
}

struct C
{
    A a; // Uses Std430DataLayout
    B b; // Uses CDataLayout
}
```

This approach may end up clashing with the existing per-buffer data layout approach.

## 3. Both approaches combined

This would effectively boil down to the struct attribute only setting a default layout, while pointer or buffer layouts override that layout if specified.

This could be implemented by starting with approach 1 and then adding the attribute, which would be used when the pointer or buffer layout is set to `DefaultDataLayout`.

This approach may be possible to implement without breaking changes and provides most flexibility.

juliusikkala:
Closes #36 and is very useful for implementing https://github.com/shader-slang/slang/issues/9478 cleanly.

Some points for discussion:
* Do we want to expose this functionality to users? The proposal currently assumes so, but the data layouts could also be just an IR-level thing to enable lowering of buffer types to pointers.
* Should we instead take a step back and consider annotating the data layout in the type declaration instead of as a buffer parameter? This is the alternative mentioned in the proposal, but it would be a breaking change.

### DIFF
--- a/proposals/034-pointer-data-layout.md
+++ b/proposals/034-pointer-data-layout.md
@@ -80,17 +80,25 @@ A new generic parameter should be added to the `Ptr` type, indicating the
 data layout used when accessing data through the pointer. The type parameter
 should satisfy `IBufferDataLayout` and default to `DefaultDataLayout`.
 
+A pointer type explicity using the Std430 data layout would then look like
+`Ptr<Foo, Access.ReadWrite, AddressSpace.Device, Std430DataLayout>`.
+
 Detailed Explanation
 --------------------
 
 Instead of assuming the native layout for pointers, backends should determine
-the layout from the pointer type. `DefaultDataLayout` should retain current
-behavior in order to make this change non-breaking.
+the layout from the pointer type. If the layout is `DefaultDataLayout`, the
+backends should retain their current behavior in order to make this change
+non-breaking.
 
 Support should be added on all targets that support both pointers and data
 layouts; that is, SPIR-V and LLVM. Support is excluded from C++ and CUDA targets
 because they do not support non-default data layouts. HLSL, GLSL, MSL and WGSL
 targets are excluded as they do not support pointers.
+
+The LLVM emitter currently attempts to deduce the layout of a pointer by
+tracing where it came from. That logic should be removed and replaced by just
+using the layout parameter of the pointer type.
 
 Alternatives Considered
 -----------------------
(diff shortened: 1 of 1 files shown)


csyonghe (approved):
This is probably the easiest way forward given what we have for now.

An alternative approach, although requiring a larger refactor, is to model laid-out types explicitly, so we don't need to conflate the pointer type itself with even more generic parameters.

For example, we can have:
1. Ordinary types, such as `struct Foo { int x; int y; }:`
2. Type layout types, such as `Std430<Foo>`.
3. Pointer and buffer type's element type should always be a type layout type, that is, it is invalid to have a direct `ConstantBuffer<Foo>` type, instead you need to have `ConstantBuffer<Std430<Foo>>`, at least internally. The type checker can allow loose syntax like `ConstantBuffer<Foo>` and translate it to `ConstantBuffer<DefaultConstantBufferLayout<Foo>>` under the hood.

This model will align better with SPIRV, since `Foo` and `Std140<Foo>` will simply map to different struct types in SPIRV with different offset decorations.

And it will allow us to have more natural syntax, such as:
```
ConstantBuffer<std430 Foo> cb;
scalar Foo* ptr = ...;
```

Issue #40, opened 2026-02-05
PRs and proposal statuses need updating and triage


swoods-nv:
We have a number of proposals which have sat unmoving for a long time as PRs. We additionally have a number of proposals which have moved to implementation without their status being updated in the proposal doc. Let's clean these out so it's easier to get a good view of what's in flight and what's usable.

swoods-nv:
PRs #47 through #53 merged.

PR #56, opened 2026-05-14
Add texture format types proposal

csyonghe:
﻿## Summary
- update the proposal to SP #040 and link the implementation status to shader-slang/slang#11344
- revise the design around the current `ITexelData` / `ITexelElement` split, with format descriptors naming a `DataType` instead of carrying scalar element metadata directly
- switch descriptor type names to Slang-style type names such as `RGBA16F`, `RGBA8UI`, and `R32F`
- address review feedback on naming rationale, no implicit format erasure, and why `RGBA32F` should not just be a `float4` alias

## Implementation
- shader-slang/slang#11344


proposals/000-texture-format-types.md
```diff
@@ -0,0 +1,224 @@
...
+
+* `RWTexture2D<rgba16f>` reads and writes `float4` values and has storage format
+  `rgba16f`.
+* `RWTexture2D<rgba8ui>` reads and writes `uint4` values and has storage format
+  `rgba8ui`.
+* `RWTexture2D<float4>` remains valid source code. Its storage format is
+  unspecified at the type level unless the compiler infers a default format or a
+  format attribute is used.
+* Two texture types with different nonzero format operands are distinct types,
+  even if their `DataType` is the same.
```
tangent-vector: One question is whether we should support an implicit conversion from a texture type with an explicit format to one that has the same data type (and shape, etc.) but uses the default format.

I think it is fine to not support that in a first pass, but not allowing that conversion for this new syntax/approach might risk users still preferring to use `[format(...)]` because it lets them drop format information when calling functions.

proposals/000-texture-format-types.md
```diff
@@ -0,0 +1,224 @@
...
+
+Detailed Explanation
+--------------------
+
+### Format descriptor types
+
+For each image format supported by Slang, the standard library should define a
+format descriptor type whose name matches the image format spelling, for example
+`r32f`, `rg32f`, `rgba16f`, `rgba8`, `rgba8ui`, and `r32i`.
```
tangent-vector: I hope that there will be material later in this proposal that provides a string rationale for the naming choices here. Following the most common naming convention for types in Slang, these would be `R32F`, `RG32F`, etc.

There is also the question of whether to adopt D3D-style naming for the formats or use this Vulkan-style naming.

I don't have a deeply held opinion about any of these details, but they need to at least be raised as points for consideration and decision-making.

proposals/000-texture-format-types.md
```diff
@@ -0,0 +1,224 @@
...
+
+This allows source compatibility while giving new code a more precise spelling
+when the storage format matters.
+
+For writable/storage textures, the compiler may infer conventional default
+formats from legacy data-type arguments when targeting APIs that require a
+format. For example, `RWTexture2D<float4>` may infer `rgba32f`, and
+`RWTexture2D<int4>` may infer `rgba32i`. Such inference is a compatibility
+feature. New code should prefer explicit format descriptor types whenever the
+resource format is known.
```
tangent-vector: It seems like we should discuss the possible alternative of having, e.g., `Texture2D<float4>` be equivalent to `Texture2D<rgba32f>`, by simply having `typealias rgba32f = float4`. In that model, there would be formats like `rgba32f` where the in-memory format for an element matches the in-memory layout of the corresponding type (so the format and type are identical, for our purposes) and then other formats where there isn't a directly corresponding data type.

I suspect the answer is that we kind of need format-agnostic textures to still be a supported language feature, because a lot of code explicitly wants to be "polymorphic" in texture format, so it is still important to be able to spell the format-agnostic texture types (and have the existing names like `Texture2D<float4>` still refer to those).

tangent-vector (approved):
I'd like to see my comments addressed in a future draft, but overall this proposal looks great.

This is a feature that you and I have talked about multiple times, Yong, and it is great to see it actually start to make forward progress.

proposals/000-texture-format-types.md
csyonghe: Yeah, we still need to maintain support for unknown formats. `Texture2D<float4>` behavior is kept as is for compatibility. Agree on future extension to support polymorphic texture formats.

proposals/000-texture-format-types.md
csyonghe: Renamed to `RGBA8Unorm` etc. to stay consistent with the slang naming convention.

proposals/000-texture-format-types.md
csyonghe: agreed. We should revisit and can always add support for it later.

PR #60, opened 2026-06-12
[codex] Add pack count constraint proposal

csyonghe:
﻿## Summary

Adds SP #041, a language proposal for variadic pack count constraints:

```slang
where countof(Pack) == IntExpr
```

The proposal documents the feature as an oriented pack-shape proof rather than a general integer equality system.

## Design Points Covered

- Type-pack and value-pack support.
- Exact declared-witness forwarding for nested generic bodies.
- Validation-only behavior: no inference of `N` from `countof(T)`.
- Oriented syntax only: `countof(Pack)` must be on the left side.
- Rejection of `!=`, `>=`, and RHS `countof(...)` spellings.
- The non-goal of algebraic proofing for cases such as `countof(T) + 1 == N`.
- The non-goal of symmetric witness use for `countof(T) == countof(U)`.
- AST, witness solving, signature matching, lowering, diagnostics, alternatives, and future directions.

## Validation

- `git.exe diff --cached --check` before commit.
- Proposal-only change; no compiler tests required in this repo.

### DIFF
--- a/proposals/041-variadic-pack-count-constraints.md
+++ b/proposals/041-variadic-pack-count-constraints.md
@@ -35,7 +35,47 @@ Those features make it possible to define APIs over an arbitrary number of type
 or value arguments, but they do not provide a way for an API to require an exact
 pack size that is itself generic.
 
-For example:
+Motivation
+----------
+
+A real-world example is a tensor API whose rank is tracked by a generic integer.
+The tensor wants a single `load` method that accepts exactly one integer index
+per dimension:
+
+```slang
+struct Tensor<T, let Dimension : int>
+{
+    T load<each Index>(Index indices)
+        where Index == int
+        where countof(Index) == Dimension
+    {
+        ...
+    }
+}
+```
+
+Without an exact pack-count constraint, the library has to spell a separate
+extension for each supported rank:
+
+```slang
+extension<T> Tensor<T, 1>
+{
+    T load(int i0) { ... }
+}
+
+extension<T> Tensor<T, 2>
+{
+    T load(int i0, int i1) { ... }
+}
+
+// ...and so on for every supported dimension.
+```
+
+That workaround bloats tensor libraries with repetitive overloads, expands the
+overload set the compiler must consider, and slows compilation for an API whose
+real requirement is simply "the number of indices equals the tensor dimension."
+
+A smaller standalone example has the same shape:
 
 ```slang
 void load<let N : int, each TIndex>(TIndex indices)
--- a/proposals/041-variadic-pack-count-constraints.md
+++ b/proposals/041-variadic-pack-count-constraints.md
@@ -74,6 +74,10 @@ extension<T> Tensor<T, 2>
 That workaround bloats tensor libraries with repetitive overloads, expands the
 overload set the compiler must consider, and slows compilation for an API whose
 real requirement is simply "the number of indices equals the tensor dimension."
+With a pack-count constraint, the compiler has a direct shape fact it can use
+during overload resolution, so candidates whose variadic argument pack cannot
+match the requested rank can be discarded without relying on a long list of
+rank-specific overloads.
 
 A smaller standalone example has the same shape:
 
@@ -116,10 +120,11 @@ substitution.
 Related Work
 ------------
 
-C++ templates can express similar relationships using non-type template
-parameters and template constraints, but those constraints are evaluated in the
-template instantiation model rather than Slang's first-class generic and witness
-system.
+C++ templates can express similar relationships using the `sizeof...` operator,
+non-type template parameters, and template constraints. For example, a C++
+template can require `sizeof...(Rest) + 1 == N` before accepting a variadic
+argument list. Those constraints are evaluated in the template instantiation
+model rather than Slang's first-class generic and witness system.
 
 Swift's variadic generics motivate Slang's `each`/`expand` model, but Swift's
 pack shape model is not directly applicable to Slang's existing generic value
@@ -220,8 +225,11 @@ struct Outer<let each D : int>
 
 ### Expected Count Expression
 
-The right side must be a compile-time integer expression that folds to an
-`IntVal`.
+The right side uses the same expression grammar and checking rules as an
+ordinary compile-time integer generic argument, such as the expression in
+`Foo<expr>`. This proposal does not introduce a separate expression grammar for
+constraints. The expression must check as a compile-time integer expression and
+fold to an `IntVal`.
 
 Examples:
 
@@ -230,10 +238,30 @@ where countof(T) == 3
 where countof(T) == N
 where countof(T) == countof(OuterPack)
 where countof(T) == N - 1
+where countof(T) == (N - 1)
 ```
 
-The last example is allowed only as an exact expected-count expression. The
-compiler does not derive algebraically equivalent facts from it.
+Parentheses, member references, and other expression forms are allowed exactly
+to the extent they are already allowed in compile-time integer generic argument
+positions and fold to an `IntVal`. For example, `countof(T) ==
+U<countof(T)>.Member` is valid only if that member access would already be a
+valid compile-time integer generic argument expression. A syntax form such as
+`N + {1}` is not accepted by this feature unless it is accepted by the existing
+compile-time expression grammar.
+
+The arithmetic examples above are allowed only as exact expected-count
+expressions. The compiler does not derive algebraically equivalent facts from
+them.
+
+The right side may itself mention `countof(...)`, including the same pack:
+
+```slang
+where countof(T) == countof(T)
+```
+
+Such a constraint is legal but still follows the oriented witness rule: it
+declares a proof about the left-side pack `T`; it does not create a general
+symmetric equality theorem.
 
 ### Concrete and Declared Witnesses
 
@@ -472,6 +500,28 @@ where countof(T) == N - 1
 If a user writes the direct oriented fact, the compiler can carry it as a
 declared witness. If the user writes an algebraically related fact, the compiler
 does not derive a new one.
+
+### General value-parameter constraints
+
+We also considered whether a limited value-constraint form could cover this
+feature by requiring the left side to be a generic value parameter:
+
+```slang
+where N == IntExpr
+```
+
+That direction may be useful in the future, but it does not directly model the
+proof needed here. A pack-count constraint is a fact about a pack parameter:
+`bar<N, I>()` needs a witness that the substituted pack `I` has count `N`, and
+the witness must be forwarded through generic substitution. A constraint on the
+value parameter `N` would answer a different question: it would constrain or
+define the value argument, and would still need additional rules to turn that
+fact into an oriented proof about a particular pack.
+
+Keeping the left side as `countof(Pack)` therefore keeps the proof object tied
+to the pack whose shape is being constrained. It also leaves room for a future
+general value-constraint or boolean-constraint system to accept compatible
+surface syntax without requiring this feature to define such a system now.
 
 ### Symmetric spelling
 
--- a/proposals/041-variadic-pack-count-constraints.md
+++ b/proposals/041-variadic-pack-count-constraints.md
@@ -133,6 +133,12 @@ where-clause
      |  'countof' '(' pack-param ')' '==' int-expr
 ```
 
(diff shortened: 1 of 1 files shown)


proposals/041-variadic-pack-count-constraints.md
```diff
@@ -0,0 +1,503 @@
...
+validate concrete calls but cannot carry the abstract proof through generic
+substitution.
+
+Related Work
+------------
+
+C++ templates can express similar relationships using non-type template
+parameters and template constraints, but those constraints are evaluated in the
+template instantiation model rather than Slang's first-class generic and witness
+system.
```
skiminki-nv: Just a note. This probably refers to the [sizeof... operator](https://en.cppreference.com/cpp/language/sizeof...).

Something like the below is probably the model example:
```c++
#include <cstddef>
#include <initializer_list>
#include <iostream>
#include <type_traits>

template <typename FirstArgType, typename SecondArgType, typename... RestArgTypes>
constexpr bool areSameTypes() noexcept
{
    if constexpr(sizeof...(RestArgTypes) == 0U)
        return std::is_same_v<FirstArgType, SecondArgType>;
    else
        return
            std::is_same_v<FirstArgType, SecondArgType> &&
            areSameTypes<FirstArgType, RestArgTypes...>();
}

template <size_t N, typename FirstArgType, typename... RestArgTypes>
requires (sizeof...(RestArgTypes) + 1U == N &&
          areSameTypes<FirstArgType, RestArgTypes...>())
auto sumArgs(FirstArgType arg, RestArgTypes... args) -> FirstArgType
{
    FirstArgType ret { };
    for (auto i : std::initializer_list<FirstArgType> { arg, args... })
        ret += i;

    return ret;
}

int main()
{
    // Declared argument count must match the number of arguments.
    // Argument types must be the same.
    std::cout << "Sum: " << sumArgs<3>(1.0, 2.5, 3.0) << std::endl;
}
```

proposals/041-variadic-pack-count-constraints.md
```diff
@@ -0,0 +1,503 @@
...
+-----------------
+
+### Syntax
+
+Add a generic constraint form:
+
+```text
+where-clause
+    ::= ...existing forms...
+     |  'countof' '(' pack-param ')' '==' int-expr
```
skiminki-nv, line 133: I'm not particularly fond of adding new where-clause syntax here. For the long run, better would be to allow boolean-typed constraint expressions, and if necessary, optimize expressions such as `countof(T) == N` for the equivalent overload resolution perf.

However, the current approach should be extensible to the more generic approach without breakages, if we'd ever implement one. So, this is again, more a note than anything else.

proposals/041-variadic-pack-count-constraints.md
```diff
@@ -0,0 +1,503 @@
...
+Status: Implementation In-Progress
+([shader-slang/slang#11571](https://github.com/shader-slang/slang/pull/11571)).
+
+Author: Yong He
+
+Reviewer:
+
+Background
+----------
+
```
skiminki-nv, line 28: While this proposal is useful as a generic API constraint, I would perhaps also add the concrete motivation behind this proposal, which is the overload resolution perf (IIUC).

That is, the new `where counof(...) == N` constraint allows the compiler to quickly discard a bunch of overload candidates in case the application has declared a large number of those with a differing number of arguments.

proposals/041-variadic-pack-count-constraints.md
```diff
@@ -0,0 +1,503 @@
...
+The chosen design avoids that trap by requiring the exact pack-shape fact to be
+written directly:
+
+```slang
+where countof(T) == N - 1
+```
+
+If a user writes the direct oriented fact, the compiler can carry it as a
+declared witness. If the user writes an algebraically related fact, the compiler
+does not derive a new one.
```
juliusikkala, line 527: Could general value constraints be achievable the same way, by always requiring that the left side is a reference to a generic parameter?

proposals/041-variadic-pack-count-constraints.md
```diff
@@ -0,0 +1,503 @@
...
+```slang
+load(1, 2, 3)
+```
+
+to infer `N == 3` from `where countof(TIndex) == N`.
+
+This was rejected for the initial feature. Pack-count constraints validate
+already-solved arguments; they do not create new ordinary generic argument
+constraints. That keeps generic inference predictable and avoids requiring a
+larger integer-constraint solving system.
```
juliusikkala, line 591: I think this should be rejected altogether, not only for the initial feature. One can form fifth-degree equations to make this practically unsolvable: `where countof(TIndex) == 3*N*N*N*N*N-2*N*N*N+N*N+N-432`

It can also have interesting ambiguities: `where countof(TIndex) == N+M` or `where countof(TIndex) == N*N`.

proposals/041-variadic-pack-count-constraints.md
```diff
@@ -0,0 +1,503 @@
...
+The right side must be a compile-time integer expression that folds to an
+`IntVal`.
+
+Examples:
+
+```slang
+where countof(T) == 3
+where countof(T) == N
+where countof(T) == countof(OuterPack)
+where countof(T) == N - 1
```
juliusikkala, line 230: So, _any_ integer expression is OK?

* Is `N+{1}` going to be legal?
    - I'm asking this with parsing in mind. That's not impossible to parse by any means, but could prove annoying in the long run. Particularly so, if we ever allow the C++-style `Type{initializer}` syntax.
* Can parentheses be used too, like `(N - 1)`? 
* `countof(T) == countof(T)`?
* `countof(T) == U<countof(T)>.Member`? (where `U<int N>` contains `static const int Member = N;`)

proposals/041-variadic-pack-count-constraints.md
csyonghe: This alternative, along with the reasoning to go with the current approach is discussed in "alternative solutions" section below.

proposals/041-variadic-pack-count-constraints.md
csyonghe: yes, but that is kept out of this proposal to keep the scope small.

proposals/041-variadic-pack-count-constraints.md
csyonghe: This means any expr that is already allowed in compile-time constant locations such as in `foo<expr>`.

proposals/041-variadic-pack-count-constraints.md
csyonghe: Done.

proposals/041-variadic-pack-count-constraints.md
csyonghe: Yes, I don't think we can support this any time soon, if ever.

proposals/041-variadic-pack-count-constraints.md
```diff
@@ -0,0 +1,577 @@
...
+{}
+
+void badShape<let N : int, each T>()
+    where countof(__trimFirst(T)) == N // Error in this proposal.
+{}
+```
+
+Nested generic declarations must constrain their own pack parameters. A nested
+generic may refer to an outer pack in the expected count expression, but the
+left-side pack being constrained must belong to the current generic:
```
tangent-vector, line 199: Unrelated to this specific proposal, but allowing nested generics (or nested declarations *in general*) to include `where` clauses that place additional constraints on outer generic parameters is a nice-to-have feature with clear semantics ("this member/method is only usable when the client can prove that these additional constraints are satisfied"). It wouldn't express anything that couldn't be done via an `extension`, but might still be a convenient syntax.

proposals/041-variadic-pack-count-constraints.md
```diff
@@ -0,0 +1,577 @@
...
+
+A pack-count constraint lowers as a hidden witness parameter, like other generic
+constraint witnesses.
+
+Declared witnesses lower through the substitution environment. Concrete
+witnesses are proof-only values: they exist so generic argument lists and
+substitution environments line up, but they do not require a runtime operation.
+
+The IR does not need to encode a general integer equality solver. It only needs
+to preserve the witness operand associated with the exact source constraint.
```
tangent-vector, line 433: Again, not actually related to this PR, but it is worth stating as a general principle: the front-end semantic checking step is responsible for generating all of the concrete proofs as well as the code that manipulates and constructs proofs (which then gets lowered to the IR). If we end up including a more general proof assistant into Slang at some point (as a way to try and handle fully general constraint equations), it would be integrated into the front-end only, and the IR passes and compiler backend would remain deterministic and (comparatively) simple.

proposals/041-variadic-pack-count-constraints.md
tangent-vector: I strongly agree with keeping this as constrained as possible to begin with, but it is probably worth noting that this proposal is syntactically compatible with the possibility of us allowing arbitrary relational expressions (or perhaps even arbitrary boolean expressions) as `where` clauses. The restrictions being described limit the scope of what is currently permitted, but not of what the language syntax can express.

That detail sets a general precedent for what we should aspire to going forward: new `where` clause cases should ideally all fit the model of `where <expr>`, and not introduce syntax that would make it harder to parse that general case down the line.

proposals/041-variadic-pack-count-constraints.md
tangent-vector: I would note that we should make a distinction between "may" and "must" here. We could say that an implementation of Slang *may* infer `N == 3` from the example given, but not that it *must* do so.

At some point, the road we are heading down somewhat necessarily involves inference becoming best-effort in certain complex scenarios, and spelling out exactly when the inference will succeed or not would be akin to spelling out under what conditions a theorem prover is guaranteed to find a solution.

We are still in the world of spelling out cases where the compiler *must* be able to perform inference, but I think we will soon need to accept that there is going to be the middle ground of things that we cannot promise the compiler will or *won't* infer.

proposals/041-variadic-pack-count-constraints.md
```diff
@@ -0,0 +1,577 @@
...
+  value pack parameter,
+- the expected count expression does not fold to a compile-time integer value,
+- a comparison operator other than `==` is used,
+- `countof(...)` appears on the right side of a comparison,
+- a concrete specialization provides a pack whose count differs from the
+  expected count,
+- an abstract specialization cannot prove the required pack-count witness.
+
+Alternatives Considered
+-----------------------
```
tangent-vector, line 449: Having read through the material here, I think the main alternative I would have liked to see discussed that isn't covered here is the option of making other changes to the language so that instead of expressing a *constraint* on two separately declared things (e.g., an integer generic parameter `N` and a variadic type pack parameter `Indices`) one instead derives one part from the other, or both from a single source of truth.

As an example, a simplistic idea might be a way to represent a type pack parameter that either takes a specific arity (an integer constant). E.g.:


```slang
struct Tensor<T, let Rank : int>
{
    T load<each[Rank] Index>(Index indices)
        where Index == int
    {
        ...
    }
}
```

Here I'm making up the notation `each[...]` to represent a type parameter pack with a pre-determined arity. In practice, that would be semantically comparable to a `where countof(Index) == Rank`, but also more directly encodes one of the limitations that this proposal relies on: the arity constraint can only apply directly to a pack parameter (always declared with `each`).

The same basic feature could be used to describe cases where the arity of an `each` pack parameter should be exactly that of another parameter via `each[countof(OtherParameter)]`.

Under this model, a declaration like the following, given earlier in the proposal:

```slang
void foo<let N : int, each I>()
    where countof(I) == N
{ ... }
```

would instead become:

```slang
void foo<let N : int, each[N] I>()
{ ... }
```

This perhaps even models what a pack parameter really is "under the hood" more directly, since it reflects how a pack parameter is effectively a parameter of type `Type[]` - that is, an array of some unknown number of types - and this notation is just allowing a parameter to be `Type[N]` - an array of a known number of types.

On balance, I think the proposal here is a good one to follow through on, but I would like to see the alternative at least discussed and hear the rationale for why we shouldn't do something along these lines.

tangent-vector (approved):
I think the proposal looks good as-is. I have one place where I made a request for more detail on alternatives, but I consider that a nice-to-have request rather than a hard requirement before we can merge this.

saipraveenb25 (approved):

PR #61, opened 2026-08-14
Propose limited HLSL templates

tangent-vector:
## Summary

This proposal adds a deliberately limited form of HLSL/C++-spelled `template` declaration to the HLSL-flavored dialect implemented by Slang.

The proposed feature provides unconstrained, duck-typed instantiation while retaining Slang's existing lookup and overload-resolution model. It defines template argument inference, SFINAE-like candidate rejection, instance caching and checking, diagnostic handling, and compiler representation. It intentionally does not add templates to the Slang language itself or attempt to implement C++ template specialization, argument-dependent lookup, or lazy member instantiation.

## Motivation

Existing HLSL codebases increasingly use `template<...>` for straightforward parametric code. Supporting that use case in the HLSL compatibility dialect can reduce migration friction without making C++ compatibility a design goal for Slang.

## Implementation note

The proposal leaves the precise rejection point for cross-module template use to an implementation spike, while defining same-module-only instantiation as the language rule.

## Tracking issue

Implementation is tracked by shader-slang/slang#12550:
https://github.com/shader-slang/slang/issues/12550

### DIFF
--- a/proposals/000-limited-hlsl-templates.md
+++ b/proposals/000-limited-hlsl-templates.md
@@ -1,0 +1,839 @@
+SP #000: Limited HLSL Templates
+===============================
+
+This proposal extends the HLSL-flavored dialect implemented by the Slang compiler with a deliberately limited form of `template` declaration.
+
+A *limited HLSL template* is a parametric declaration whose body is checked after template arguments have been supplied.
+It provides familiar HLSL/C++-style `template<...>` spelling and permits unconstrained, duck-typed bodies without requiring the author to first define Slang interfaces.
+
+The templates proposed here are not C++ templates.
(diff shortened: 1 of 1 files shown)


tangent-vector:
We are closing this PR and withdrawing the proposal at this time.

The decision to not pursue template support in the Slang toolset at this time comes down to a few overlapping factors:

- The design and architecture of the Slang language and compiler would make full compatibility with templates as they appear in C++ or current HLSL prohibitively difficult to implement.

- Most of the use cases that the "limited" templates proposed here would cover are use cases where Slang's generics are also applicable. Effort spent on implementing limited templates could instead be spent on improving the ergonomics/UX of generics.

We encourage users to try using Slang generics in those cases where they currently want to use templates, and to report issue and feature requests based on their experience so that we can continue to make generics better and more usable.

